PR #80, opened 2015-04-15
Add dedicated 'stig' and 'scap' baseline states

lorengordon:
Makes it easier to include a specific baseline.

### DIFF
--- a/ash-linux/scap.sls
+++ b/ash-linux/scap.sls
@@ -1,0 +1,2 @@
+include:
+  - ash-linux.SCAPonly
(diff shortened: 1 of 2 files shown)


ferricoxide:
Remember that while STIG does not contain 100% of SCAP, SCAP contains 100% of STIG. The two trees, as authored, have the SCAP components defined as "just the elements not already taken care of in the STIG components"

lorengordon:
Right. I'll add that back to the scap baseline.

lorengordon:
Done in https://github.com/lorengordon/ash-linux-formula/commit/c8cabea9a1445ebedf403f79cd22fb37ca952442.

PR #94, opened 2015-10-19
Warn, but do not exit with failure

lorengordon:


ferricoxide:
Might have been better to change the `exit 1` to `> /dev/stderr`. Still allows things to move forward, but outputs the error-message more appropriately.

lorengordon:
I don't think salt differentiates, really. Either way, it's reported as a
"comment" in the state output. Not worth spending the time to think about
it.

PR #145, opened 2016-05-25
V38513: Manage iptables safety rules in ASH chain

### LINKED ISSUE #144: System becomes unreachable if `/etc/sysconfig/iptables` exists prior to running the ash-linux formula
Due to the patch that fixed setting the INPUT table to DROP, and the jinja `if` test wrapping the application of rules allowing ssh and established,related, if the `/etc/sysconfig/iptables` file exists prior to running the ash-linux formula the system will become unreachable.

See https://github.com/plus3it/SystemPrep/issues/90.

lorengordon:
Previously, V38513.sls used a jinja test to check for the existence
of `/etc/sysconfig/iptables`, and only created the safety rules for
iptables if the file was missing. However, the `INPUT` table was
_always_ being configured to `DROP`. For systems where this file
already existed, this combination caused the system to become
unreachable because the safety rules were not being added.

This patch updates V38513.sls to use the newer functions and
parameters available in the `iptables` state module to create a new
chain, `ASH`, jump from the `INPUT` chain to the `ASH` chain, and
creates the safety rules in the `ASH` chain.

Creating the saftey rules is conditional on whether the jump rule
is added to the `INPUT` chain. This means users can modify the
`ASH` chain after the fact if they wish to change the default safety
rules.

Fixes #144

### DIFF
--- a/ash-linux/el6/STIGbyID/cat2/V38513.sls
+++ b/ash-linux/el6/STIGbyID/cat2/V38513.sls
@@ -3,10 +3,10 @@
 # Version:	RHEL-06-000120
 # Finding Level:	Medium
 #
-#     The systems local IPv4 firewall must implement a deny-all, 
-#     allow-by-exception policy for inbound packets. In "iptables" the 
-#     default policy is applied only after all the applicable rules in the 
-#     table are examined for a match. Setting the default policy to "DROP" 
+#     The systems local IPv4 firewall must implement a deny-all,
+#     allow-by-exception policy for inbound packets. In "iptables" the
+#     default policy is applied only after all the applicable rules in the
+#     table are examined for a match. Setting the default policy to "DROP"
 #     implements proper design for a firewall, ...
 #
 #  CCI: CCI-000066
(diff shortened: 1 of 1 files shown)


ferricoxide:
Described methodology looks sound enough. Assuming it works as advertised across AMIs, should be good to go.

ferricoxide:
Looking at code: it should be noted that the jump to ASH will always need to be the _**first**_ rule in the `INPUT` chain. If tenants subsequently insert other rules ahead of the `INPUT` chain's jump to the ASH chain, performance (under high network loads) will be adversely impacted as the `ESTABLISHED,RELATED` rule will end up processing much later than when it's the very-first rule in the `INPUT` chain

lorengordon:
We could use `iptables.insert` instead of `iptables.append` to specify a location in the chain, but then we run into problems where the user happens to change both the location of the jump rule in the `INPUT` chain _and_ the default safety rules in the `ASH` chain. In that scenario, the jump rule would be moved to be first, which would trigger the `onchanges` requisite on the safety rules and overwrite whatever changes they made to the `ASH` chain...

ferricoxide:
Mostly providing input as it relates to performance-related ordering.

Primary expectation for the `ASH` chain was that it would contain _only_ the SSH rule. Was mostly expecting that the "standard", performance-oriented rules would remain in the `INPUT` chain. If the only thing in the `ASH` chain is the SSH rule, it doesn't matter whether the jump-to-`ASH` is rule 1 or rule 1,000.

ferricoxide:
Similarly, if the `ESTABLISHED,RELATED` rule is redundant — in both the standard `INPUT` chain and the custom `ASH` chain, whichever one matches first will be processed first. So, you could be ~~paranoid~~ _super_-safe and make the rule present in both.

Or, one can take the stance "this is for compliance not optimization: tenants that want/need optimized performance can insert the performance-enhancing rules as the see fit (even if it creates redundancies with rules in other chains).

/me shrugs

lorengordon:
Ok, how about this logic:

- If the jump rule is not present, append it (`iptables.append`)
- If the jump rule is appended, insert the established,related rule to `INPUT` at position 1 (`iptables.insert`)
- If the jump rule is appended, append the ssh safety rule (`iptables.append`)

With this logic, if they change the safety rule, no problem. If they change the position of the jump rule, no problem. If they remove the jump rule entirely, _and_ change the position of the establish,related rule, both will be re-applied. I think this is the most reasonable approach I can come up with...

ferricoxide:
Sounds like a plan, Stan.

lorengordon:
Ok, testing that approach now...

lorengordon:
Also fixes https://github.com/plus3it/SystemPrep/issues/90

PR #194, opened 2016-12-26
Lists iptables before saving rules

lorengordon:
Salt 2016.11 runs `iptables-save` when calling `iptables.save`, and
then writes the output to `/etc/sysconfig/iptables`. However, in
some cases, `iptables-save` appears to generate no output. First
running `iptables --list` causes `iptables-save` to output the
expected rules in all (known) cases.


lorengordon:
The error also occurs on salt 2016.3.

PR #197, opened 2016-12-26
Uses `replace` with file.managed only when a source is set

### LINKED ISSUE #40: Handler for V-38679 should call out IPs that don't have boot-time config files
While not expressly required by the STIGs, should probably iterate _all_ the IPs/interfaces and ensure they have boot-time configuration files.

lorengordon:
Fixes #40


lorengordon:
Correct, though note that I did _not_ remove the `replace` token if `source`, `contents`, or `contents_pillar` was specified. There was one instance that matched that condition.

PR #198, opened 2016-12-26
Eliminates deprecation warning by using cmd.shell

### LINKED ISSUE #193: Some sls is using `cmd.run` in Jinja without `python_shell`
This alias mentioned in the deprecation warning is getting removed in the next version of salt, so we need to find these and fix them up.

Unfortunately, the error does not indicate _which_ sls contains the deprecated usage.

```
/usr/lib/python2.6/site-packages/salt/utils/templates.py:73: DeprecationWarning: Starting in 
2015.5, cmd.run uses python_shell=False by default, which doesn't support shellisms (pipes, 
env variables, etc). cmd.run is currently aliased to cmd.shell to prevent breakage. Please 
switch to cmd.shell or set python_shell=True to avoid breakage in the future, when this 
aliasing is removed.
```

lorengordon:
Fixes #193


lorengordon:
Yep, more or less (since I used the atom editor to make the change).

PR #209, opened 2017-02-09
Handlers for /tmp mount options

ferricoxide:
Vendor auto-remediation doesn't know how to remediate when /tmp is systemd-managed (as required for tmpfs-based /tmp under EL7)

### DIFF
--- a/ash-linux/el7/VendorSTIG/remediate.sls
+++ b/ash-linux/el7/VendorSTIG/remediate.sls
@@ -5,7 +5,11 @@
 #################################################################
 {%- set stig_id = 'VendorSTIG-top' %}
 {%- set helperLoc = 'ash-linux-formula/ash-linux/el7/VendorSTIG/files' %}
-{%- set dsos = salt.grains.get('os')|lower %}
+{%- if salt.grains.get('os')|lower == 'redhat' %}
+  {%- set dsos = 'rhel' %}
+{%- else %}
+  {%- set dsos = salt.grains.get('os')|lower %}
+{%- endif %}
 {%- set osrel = salt.grains.get('osmajorrelease') %}
 {%- set contentDir = '/usr/share/xml/scap/ssg/content' %}
 {%- set dsfile = 'ssg-' + dsos + osrel + '-ds.xml' %}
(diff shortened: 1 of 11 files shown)


ash-linux/el7/VendorSTIG/remediate.sls
```diff
@@ -5,7 +5,11 @@
 #################################################################
 {%- set stig_id = 'VendorSTIG-top' %}
 {%- set helperLoc = 'ash-linux-formula/ash-linux/el7/VendorSTIG/files' %}
-{%- set dsos = salt.grains.get('os')|lower %}
+{%- if salt.grains.get('os')|lower == 'redhat' %}
+  {%- set dsos = 'rhel' %}
```
lorengordon: This set is scoped to the if block. Outside the if block, the var is out of scope. Might want to look at another approach. Perhaps an inline if...

`{%- set dsos = 'rhel' if salt.grains.get('os')|lower == 'redhat' else salt.grains.get('os')|lower %}`

ash-linux/el7/VendorSTIG/report.sls
```diff
@@ -6,7 +6,11 @@
 {%- set stig_id = 'VendorSTIG-top' %}
 {%- set helperLoc = 'ash-linux-formula/ash-linux/el7/VendorSTIG/files' %}
 {%- set repDir = '/var/tmp' %}
-{%- set dsos = salt.grains.get('os')|lower %}
+{%- if salt.grains.get('os')|lower == 'redhat' %}
+  {%- set dsos = 'rhel' %}
```
lorengordon: Same scope issue.

ash-linux/el7/VendorSTIG/remediate.sls
ferricoxide: Weird... It works at setting the string-var globally. Only place I've had scoping issues is with complex var-types.

ash-linux/el7/VendorSTIG/report.sls
ferricoxide: (Insert equivalent response)

ash-linux/el7/VendorSTIG/remediate.sls
lorengordon: I suppose if it works...

ferricoxide: Dunno. Ive seen scope discussions around lists and bools, but not simple strings (even when my search parameter includes "string")

PR #215, opened 2017-03-10
Issue 214

### LINKED ISSUE #214: Nessus scans reporting on weak MACs and CBC ciphers used by sshd
Potential fix appears to be https://access.redhat.com/solutions/420283

ferricoxide:
Fixes #214 
* Created a Nessus hierarchy under el6 (may move  upwards if there's a high degree of El6/7 overlap)
* Created a `MACs`-handler
* Created a `Ciphers`-handler
* Deprecated stale STIG-handler fo V38617 (didn't want collisions with new Nessus-handlers)


ash-linux/iavm.sls
```diff
@@ -0,0 +1,5 @@
+{%- set ver = grains['osmajorrelease'] %}
+include:
+  - ash-linux.stig
+  - ash-linux.el{{ ver }}.SCAPonly
+  - ash-linux.el{{ ver }}.Nessus
```
lorengordon: I think this should probably only include the Nessus line. The others would get included through `ash-linux/init.sls`. That way we could apply _just_ the iavm baseline on its own, if we needed to.

ferricoxide: Yeah, I was iffy on this one ...but didn't want to hear "why didn't you link it in so it runs?" =)

lorengordon: Heh, I see similar in the scap baseline now. Yeah, can probably re-scope that one too. :p

PR #227, opened 2017-07-05
Changing FIPS Mode

KevinPlus3:
Updated _modules to enable/disable FIPS for a Linux system following the steps provided in documentation.  Also checks status of FIPs for the system.

Adds a _states to enable/disable FIPS.

### DIFF
--- a/_modules/ash_linux.py
+++ b/_modules/ash_linux.py
@@ -20,28 +20,51 @@ def __virtual__():
         return False, 'ash_linux module works only on Linux systems'
 
 
-def _create_new_grub_file(remove):
+def _move_boot_kernel(restore_bak):
+    """
+    Create/restore backup of boot-kernel file.
+
+    Args:
+        restore_bak:  (`obj`: `bool`)
+        True if restoring backup of boot-kernel. False to create a backup.
+    """
+    if restore_bak:
+        filename = "initramfs-" + os.uname()[2] + ".img.bak"
+        new_path = os.path.join("/boot", filename[:-4])
+    else:
+        filename = "initramfs-" + os.uname()[2] + ".img"
+        new_path = os.path.join("/boot", filename + ".bak")
+
+    old_path = os.path.join("/boot", filename)
+    if os.path.exists(old_path):
+        __salt__['file.move'](old_path, new_path)
+
+    return old_path
+
+
+def _modify_grub_file(rmv_fips_arg):
     """
     Insert/remove fips argument from GRUB file in /etc/default.
     
     Args:
-        remove:  (`obj`: `bool`)
+        rmv_fips_arg:  (`obj`: `bool`)
         True if removing fips argument. False to add it.
     """
-    if os.path.exists('/etc/default/grub'):
-        with salt.utils.fopen('/etc/default/grub') as old_fle:
-            with salt.utils.fopen('/etc/default/grub_new', 'w') as new_fle:
-                for line in old_fle.xreadlines():
-                    if line.startswith('GRUB_CMDLINE_LINUX') and remove:
-                        new_fle.write(line.replace('fips=1 ', ''))
-                    elif line.startswith('GRUB_CMDLINE_LINUX'):
-                        new_fle.write(line.replace('boot=', 'fips=1 boot='))
-                    else:
-                        new_fle.write(line)
-        __salt__['file.move']('/etc/default/grub_new', '/etc/default/grub')
+    filepath = '/etc/default/grub'
+    if rmv_fips_arg:
+        result = __salt__['file.replace'](
+            filepath, 'fips=1', '', show_changes=False
+        )
+    else:
+        result = __salt__['file.replace'](
+            filepath, 'GRUB_CMDLINE_LINUX="',
+            'GRUB_CMDLINE_LINUX="fips=1 ',
+            show_changes=False
+        )
+    return result
 
 
-def _get_dracutfips_pkgs():
+def _get_installed_dracutfips_pkgs():
     """
     Returns list of available dracut-fips pkgs to install. We currently
     resort to using an actual call to yum as Salt's pkg.list_repo_pkgs()
@@ -49,9 +72,36 @@ def _get_dracutfips_pkgs():
     instead of "enabled=1".  Once that has been addressed in Salt, we can
     update this to use Salt's command.
     """
-    cmd = "yum list available"
+    cmd = "yum list installed"
     available_pkgs = __salt__['cmd.run'](cmd, python_shell=False)
     return re.findall(r'\bdracut-fips[^.]*', available_pkgs)
+
+
+def _get_grub_args():
+    cmd = "grubby --info=ALL | grep args="
+    return __salt__['cmd.run'](cmd, python_shell=True)
+
+
+def _rollback_fips_disable(installed_fips_pkgs):
+    """
+    Rollback the actions of fips_disable() upon a thrown error.
+
+    Args:
+        installed_fips_pkgs:  (`obj`: `list`)
+        List of installed dracut-fips packages that were removed
+        during the process of running fips_disable().
+    """
+    __salt__['pkg.install'](installed_fips_pkgs)
+    _move_boot_kernel(True)
+
+    grub_bak = '/etc/default/grub.bak'
+    if os.path.exists(grub_bak):
+        __salt__['file.move'](grub_bak, '/etc/default/grub')
+
+    __salt__['cmd.run'](
+        "grubby --update-kernel=ALL --args=fips=1",
+        python_shell=False
+    )
 
 
 def fips_disable():
@@ -64,31 +114,62 @@ def fips_disable():
     .. code-block:: bash
         salt '*' ash.fips_disable
     """
-
-    # Remove dracut-fips installations.
-    __salt__['pkg.remove']('dracut-fips')
-
-    # Create a back-up of the FIPS boot-kernel.
-    # The third index in the os.uname() tuple contains the release information.
-    filename = "initramfs-" + os.uname()[2] + ".img"
-    old_path = os.path.join("/boot", filename)
-    new_path = os.path.join("/boot", filename + ".FIPS-bak")
-    __salt__['file.move'](old_path, new_path)
-
-    # Create a new boot-kernel.
-    __salt__['cmd.run']("dracut -v", python_shell=False)
-
-    # Update grub.cfg file to remove the fips argument.
-    __salt__['cmd.run']("grubby --update-kernel=ALL --remove-args=fips=1")
-
-    # Update GRUB command line entry to remove fips.
-    _create_new_grub_file(True)
-
+    installed_fips_pkgs = _get_installed_dracutfips_pkgs()
     ret = {
         'result': True,
-        'comment': "FIPS set to disable. Reboot system to disable FIPS."
+        'changes': {'old': {}, 'new': {}},
+        'comment': ("FIPS has been toggled to off.",
+                    "Reboot system to place into FIPS-disabled state.")
     }
-    return ret
+
+    try:
+        # Remove dracut-fips installations.
+        installed_fips_pkgs = _get_installed_dracutfips_pkgs()
+        if 'dracut-fips' in installed_fips_pkgs:
+            __salt__['pkg.remove']('dracut-fips')
+            ret['changes']['old'].update({'Packages': installed_fips_pkgs})
+
+        # Create a back-up of the FIPS boot-kernel.
+        # The third index in the os.uname() tuple contains the release information.
+        _move_boot_kernel(False)
+
+        # Create a new boot-kernel.
+        __salt__['cmd.run']("dracut -f", python_shell=False)
+
+        # Update grub.cfg file to remove the fips argument.
+        grub_args = _get_grub_args()
+        if 'fips=1' in grub_args:
+            cmd = 'grubby --update-kernel=ALL --remove-args=fips=1'
+            __salt__['cmd.run'](cmd, python_shell=False)
+            ret['changes']['new'].update({'grubby': cmd})
+
+        # Update GRUB command line entry to remove fips.
+        if _modify_grub_file(True):
+            ret['changes']['old'].update({'/etc/default/grub': {
+                'GRUB_CMDLINE_LINUX': 'fips=1'
+            }})
+    except:
+        _rollback_fips_disable(installed_fips_pkgs)
+        ret['result'] = False
+        ret['changes'] = {}
+        ret['comment'] = 'Unable to change state of system to FIPS-disabled.'
+    finally:
+        return ret
+
+
+def _rollback_fips_enable():
+    """Rollback the actions of fips_enable() upon a thrown error."""
+    __salt__['pkg.remove']('dracut-fips')
+    _move_boot_kernel(True)
+
+    grub_bak = '/etc/default/grub.bak'
+    if os.path.exists(grub_bak):
+        __salt__['file.move'](grub_bak, '/etc/default/grub')
+
+    __salt__['cmd.run'](
+        "grubby --update-kernel=ALL --remove-args=fips=1",
+        python_shell=False
+    )
 
 
 def fips_enable():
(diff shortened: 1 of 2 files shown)


_states/ash_linux.py
```diff
@@ -0,0 +1,36 @@
+def _error(ret, err_msg):
+    ret['result'] = False
+    ret['comment'] = err_msg
+    return ret
+
+
+def fips_managed(status):
```
lorengordon: Would like to add support for `test=True`, which is a standard option injected into all states in the `__opts__` dictionary. The idea is `if test`, then set the comment to `'FIPS is set to be {0}'.format(status)` and return; do not actually execute the change.

_modules/ash_linux.py
```diff
@@ -4,16 +4,151 @@
...
+        salt '*' ash.fips_status
+    """
+    try:
+        with salt.utils.fopen('/proc/sys/crypto/fips_enabled', 'r') as fle:
+            val = fle.read().strip()
+            if val == '1':
+                return 'enabled'
+            else:
+                return 'disabled'
+    except:
```
lorengordon: This should probably be `except (IOError, FileNotFoundError):`. Those ought to cover py2 and py3, I think.

_states/ash_linux.py
```diff
@@ -0,0 +1,36 @@
...
+    '''
+    Prepare a system to be FIPS-enabled or FIPS-disabled.
+
+    status
```
lorengordon: I think `state` might make more sense for this argument name. State seems more like the user specifying the desired config (which is how this salt state's are used). Status seems more like a question, what is the status?

_states/ash_linux.py
```diff
@@ -0,0 +1,36 @@
...
+
+    if status == 'enabled':
+        __salt__['fips.fips_enable']()
+        ret['changes'] = 'System is now FIPS-enabled.'
+        ret['comment'] = 'Reboot system to enable FIPS.'
+    elif status == 'disabled':
+        __salt__['fips.fips_disable']()
+        ret['changes'] = 'System is now FIPS-disabled.'
+        ret['comment'] = 'Reboot system to disable FIPS.'
+    else:
```
lorengordon: We should make sure this isn't a reachable code path and remove the dead-end else block.

_states/ash_linux.py
```diff
@@ -0,0 +1,36 @@
...
+           'comment': '',
+           'status': status,
+           'result': True}
+
+    current_status = __salt__['fips.fips_status']()
+    if current_status == status:
+        return ret
+
+    if status == 'enabled':
+        __salt__['fips.fips_enable']()
```
lorengordon: Can fips_enable return `False` or throw an exception? If so, we should wrap this in try/except and also capture/test the return value, before assuming successful execution.

_states/ash_linux.py
```diff
@@ -0,0 +1,36 @@
...
+        ret['changes'] = 'System is now FIPS-enabled.'
+        ret['comment'] = 'Reboot system to enable FIPS.'
+    elif status == 'disabled':
+        __salt__['fips.fips_disable']()
```
lorengordon: Same issue with assuming successful execution.

_states/ash_linux.py
```diff
@@ -0,0 +1,36 @@
+def _error(ret, err_msg):
```
lorengordon: This can be removed after ditching the else block in `fips_managed()`.

_modules/ash_linux.py
```diff
@@ -4,16 +4,151 @@
...
+    system in order for FIPS to be disabled.  This routine prepares
+    the system to disable FIPS.
+    
+    CLI Example:
+    .. code-block:: bash
+        salt '*' ash.fips_disable
+    """
+
+    # Remove dracut-fips installations.
+    __salt__['pkg.remove']('dracut-fips')
```
lorengordon: I feel like we got the pkgs approaches backwards... `fips_disable()` should remove all packages that match the pattern `"dracut-fips*"`. `fips_enable()` only needs to install `dracut-fips`.

_modules/ash_linux.py
```diff
@@ -4,16 +4,151 @@
...
+
+    # Update grub.cfg file to add the fips agurment.
+    __salt__['cmd.run']("grubby --update-kernel=ALL --args=fips=1")
+
+    # Update GRUB command line entry to add fips.
+    _create_new_grub_file(False)
+
+    ret = {
+        'result': True,
+        'comment': "FIPS set to enable. Reboot system to enable FIPS."
```
lorengordon: If we're going to set the comment this way, then the state should capture the return value and use the comment. And probably, we should also test the status in the execution module and set the comment accordingly so we aren't telling fibs to the user in case FIPS was enabled when they run the the function.

_modules/ash_linux.py
```diff
@@ -4,16 +4,151 @@
...
+
+def _create_new_grub_file(remove):
+    """
+    Insert/remove fips argument from GRUB file in /etc/default.
+    
+    Args:
+        remove:  (`obj`: `bool`)
+        True if removing fips argument. False to add it.
+    """
+    if os.path.exists('/etc/default/grub'):
```
lorengordon: And if this path does not exist?

_modules/ash_linux.py
```diff
@@ -4,16 +4,151 @@
...
+        with salt.utils.fopen('/etc/default/grub') as old_fle:
+            with salt.utils.fopen('/etc/default/grub_new', 'w') as new_fle:
```
lorengordon: Salt has atomic file operations functions. Might want to look into how to use them, rather than write new code.

_modules/ash_linux.py
```diff
@@ -4,16 +4,151 @@
...
+                for line in old_fle.xreadlines():
+                    if line.startswith('GRUB_CMDLINE_LINUX') and remove:
+                        new_fle.write(line.replace('fips=1 ', ''))
+                    elif line.startswith('GRUB_CMDLINE_LINUX'):
+                        new_fle.write(line.replace('boot=', 'fips=1 boot='))
```
lorengordon: What if `boot=` is not in the line?

_modules/ash_linux.py
KevinPlus3: The line above does remove all packages that match the pattern "dracut-fips*" ... I tested several times to confirm, so it does as intended.  If we do not care about trying to install all of the packages that start with "dracut-fips*", then yes, we can have it just install the one package.

lorengordon: Sort of. Pretty sure `pkg.remove` is not glob-matching without being told to do so explicitly. Instead, `dracut-fips-aesni` has a dependency on `dracut-fips`. So when you remove `dracut-fips` it automatically removes `dracut-fips-aesni`.

KevinPlus3: Gotcha.

_states/ash_linux.py
KevinPlus3: Not sure I follow ... right now - if "state" is set to some value that is not 'enabled' or disabled', then the code above is not reachable and "else" is run.  Are you saying you think it would be best to have the disable functionality run if state is something other than "disabled"?

lorengordon: Ahh, nvm, I misread the code. More confusion around calling this argument `status`. Thought this was the return from `ash.fips_status()`

_states/ash_linux.py
```diff
@@ -0,0 +1,36 @@
...
+        .. code-block:: yaml
+            fips.fips_managed:
+                - status: disabled
+    '''
+    ret = {'changes': {},
+           'comment': '',
+           'status': status,
+           'result': True}
+
+    current_status = __salt__['fips.fips_status']()
```
lorengordon: should be `ash.fips_status`

_states/ash_linux.py
lorengordon: should be `ash.fips_enable`

_states/ash_linux.py
lorengordon: should be `ash.fips_disable`

_states/ash_linux.py
KevinPlus3: Ooops.  Byproduct of testing in Salt without using the virtualname.

_states/ash_linux.py
lorengordon: Should also add some of the boilerplate in here:

```
# -*- coding: utf-8 -*-
"""
Provide custom state module for ash-linux.

:maintainer: Plus3 IT Systems
"""
__virtualname__ = 'ash'


def __virtual__():
    if __grains__.get('kernel', '') == 'Linux':
        return __virtualname__
    else:
        return False, 'ash_linux module works only on Linux systems'
```

_modules/ash_linux.py
ferricoxide: In order for a system with the `/boot` directory-tree on its own partition to work when FIPSed, the `boot=` token _must_ be present ...and, by extension, would not be present if `/boot` is on the same partition as `/`. Probably needs to be an additional conditional that checks to see if `/boot` is on its own device — if it is, then use the current `elif`' s contents, otherwise, need to key on a different token (probably the `root=`  token).

KevinPlus3: Man. How deep into the rabbit hole should we go?  Another consideration for example, what if `/etc/default/grub' has the line that starts with `GRUB_CMDLINE_LINUX_DEFAULT=` which appends to the end of a normal mode only as opposed to `GRUB_CMDLINE_LINUX` which covers normal and recovery mode.  Is it the case that `fips=1` will _only_ appear in the `GRUB_CMDLINE_LINUX`.  For RH, boot loaders are different if we are talking about processors outside of `AMD64` and `x86`.  Sounds like things can get quite complicated, quite quickly.

KevinPlus3: Or someone can modify files in `/etc/grub.d/##_`  .... Hehehe. Okay - looks to me like I can just append `fips=1` if it is missing from `GRUB_CMDLINE_LINUX` - it really shouldn't matter if `root=` or `boot=` is in there.

KevinPlus3: Although, as you pointed out - if I wish to enable FIPS on a system that has never been FIPSed and boot is in its own partition, then I _have_ to add `boot=...` to the `grub` file also.

ferricoxide: On the plus side, damned few people using Linux on bare-metal UEFI. The way I look at it, if those people are out there, they can feel free to provide a patch.

_modules/ash_linux.py
```diff
@@ -4,16 +4,196 @@
...
+        rollbacks.append(("__salt__['cmd.run'](",
+                          "grubby --update-kernel=ALL --args=fips=1",
+                          "python_shell=False)"))
+
+        # Update GRUB command line entry to remove fips.
+        _modify_grub_file(True)
+    except:
+        for cmmd in rollbacks:
+            exec(cmmd)
+        ret = {
```
lorengordon: This `ret`/`return` setup looks a bit messy. How about making this "False" statement the default, defined at the top of the function, and then modifying the dict keys in the try block upon success, and just `return ret` at the bottom?

lorengordon: Or, just move the "True" `ret` setup into the try block, and remove the `return` statement from the except block...

_modules/ash_linux.py
```diff
@@ -4,16 +4,196 @@
...
+        # Update GRUB command line entry to add fips.
+        _modify_grub_file(False)
+    except:
+        for cmmd in rollbacks:
+            exec(cmmd)
+        ret = {
+            'result': False,
+            'comment': "Unable to change state of system to FIPS-enabled."
+        }
+        return ret
```
lorengordon: Same deal as for `fips_disable`

_modules/ash_linux.py
lorengordon: Linters will generally complain about defining a var (`ret`) inside a try/except block, and then using it outside the block (`return ret`), so i might lean towards the first option...

KevinPlus3: (thumbsup)

_modules/ash_linux.py
```diff
@@ -4,16 +4,194 @@
...
+    ret = {
+        'result': True,
+        'changes': 'System is now in a FIPS-disabled state.',
+        'comment': "Reboot system to disable FIPS."
+    }
+
+    try:
+        # Remove dracut-fips installations.
+        __salt__['pkg.remove']('dracut-fips')
+        rollbacks.append("__salt__['pkg.install'](installed_fips_pkgs)")
```
lorengordon: Probably need to move these append statement _in front_ of the command they rollback, otherwise they won't be reached if there is an exception.

_modules/ash_linux.py
```diff
@@ -4,16 +4,194 @@
...
 
 def __virtual__():
     if __grains__.get('kernel', '') == 'Linux':
         return __virtualname__
     else:
-        return False, 'ash_linux module works only on linux systems'
+        return False, 'ash_linux module works only on Linux systems'
+
+
+def _move_boot_kernal(restore_bak):
```
lorengordon: "kernel" not "kernal"... and don't forget to update the calls to the function...

_modules/ash_linux.py
KevinPlus3: I would've thought behind ... if `__salt__` failed to package remove, then I would've thought the package would not have been removed.  Similarly for saying modifying the file, we don't want to end up "replacing" a phrase in the file if it has not been removed ... then we would end up with two of the same thing.

_modules/ash_linux.py
KevinPlus3: Dammit, I thought I caught all of them.

_modules/ash_linux.py
KevinPlus3: In addition, if one of the commands did throw an error, then we don't want to run the same command because then the error will most likely just be thrown again - we should only run commands we know has succeeded.

lorengordon: Good points, but makes me hesitate on this approach. On any failure, might just need to re-check all the conditions. :/

lorengordon: Or, first backup or store the current state, and on any failure, restore them all.

KevinPlus3: Okay, yeah.  I think I will have to go with your last suggestion.  That would be a good one.

_states/ash_linux.py
```diff
@@ -0,0 +1,48 @@
...
+        return ret
+
+    if state == 'enabled':
+        ret = __salt__['ash.fips_enable']()
+    elif state == 'disabled':
+        ret = __salt__['ash.fips_disable']()
+    else:
+        ret['comment'] = 'State {0} is not a valid option'.format(state)
+
+    return ret
```
lorengordon: nit: add an empty line

_states/ash_linux.py
```diff
@@ -0,0 +1,58 @@
...
+        Status for FIPS - either 'enabled' or 'dsiabled'.
+        For example:
+        .. code-block:: yaml
+            ash.fips_state:
+                - name: disabled
+    '''
+    state = value or name
+
+    if state == name:
+        name = 'ash.fips_state'
```
lorengordon: I don't think we should override `name` like this. It has a special meaning to salt.

_states/ash_linux.py
```diff
@@ -0,0 +1,58 @@
...
+        return __virtualname__
+    else:
+        return False, 'ash_linux module works only on Linux systems'
+
+
+def fips_state(name, value=None):
+    '''
+    Prepare a system to be FIPS-enabled or FIPS-disabled.
+
+    status
```
lorengordon: Renamed to `value`?

_states/ash_linux.py
```diff
@@ -0,0 +1,58 @@
...
+        Status for FIPS - either 'enabled' or 'dsiabled'.
+        For example:
+        .. code-block:: yaml
+            ash.fips_state:
+                - name: disabled
+    '''
+    state = value or name
```
lorengordon: Probably shouldn't override `state` either... 😟

_states/ash_linux.py
```diff
@@ -0,0 +1,58 @@
...
+
+def fips_state(name, value=None):
+    '''
+    Prepare a system to be FIPS-enabled or FIPS-disabled.
+
+    status
+        Status for FIPS - either 'enabled' or 'dsiabled'.
+        For example:
+        .. code-block:: yaml
+            ash.fips_state:
```
lorengordon: This example isn't quite complete...a state needs a name:
```
Ensure FIPS is disabled:  # <-- this is `name`
  ash.fips_state:
    - value: disabled
```
or
```
disabled:
  ash.fips_state: []
```
or this probably also works:
```
Ensure FIPS is disabled:
  ash.fips_state:
    - name: disabled  # <-- salt overrides `name`

File: _states/ash_linux.py (hunk 19 above)
Comment (KevinPlus3) [inline _states/ash_linux.py]: It is a good point. But the way we had agreed before was to use `name` for `fips_state` unless `value` was populated.  So in order to avoid having a `enabled` or `disabled` value that could potentially be in one of two variables, I think we need to just stick with one variable being the state.  I know other Salt modules use `name` as their path or value for modifying the system even though `name` may be a special meaning to Salt.  (Which leads me to believe Salt hasn't really clarified what exactly is special about `name`).  So I suggest we either do away with `value` and say this is what `name` means in t his context or say `name` represents something else and that `value` must be populated.

Comment (lorengordon) [inline _states/ash_linux.py]: Salt is perfectly clear about how they use `name`. We are the ones who are unclear. :)

I don't find it confusing to say that `value` takes precedence if it is set, but I don't really care as long as we do not override `name` or `state`.

Comment (KevinPlus3) [inline _states/ash_linux.py]: Not so sure about Salt being perfectly clear.  ;-) I haven't yet found documentation on what exactly `name` is representative of in Python.

But I agree with not overriding `name` or `state` since they do seem to have special meaning to Salt.  However, I definitely don't think it is good practice to have two variable that could potentially mean the same thing.  So I will force the method to use `value` to set the state for FIPS enablement or disablement. `name` can be used however the end user sees fit.

Comment (lorengordon) [inline _states/ash_linux.py]: https://docs.saltstack.com/en/latest/ref/states/highstate.html#name-declaration

Comment (lorengordon) [inline _states/ash_linux.py]: also described a bit in this doc, though most clearly in the return value rather than as an input, https://docs.saltstack.com/en/latest/ref/states/writing.html#states-are-easy-to-write

Comment (KevinPlus3) [inline _states/ash_linux.py]: Wooootz. I like making you find stuff for me.  :P

File: _states/ash_linux.py [hunk 23]
```diff
@@ -0,0 +1,57 @@
...
+
+    if __opts__['test']:
+        ret['comment'] = 'System is now FIPS-{0}.'.format(value)
+        return ret
+
+    current_state = __salt__['ash.fips_status']()
+    if current_state == value:
+        return ret
+
+    if not value:
```
lorengordon: This can't happen, since `value` has no default. Python itself will throw an exception first, I think... Also, the canonical thing to do here would be to raise a SaltInvocationError...

KevinPlus3: Would a `SaltInvocationError` also be the proper thing to do for if `value` is something other than `enabled` or `disabled`?

lorengordon: Yep. It's neat, it ought to print out the doctstring for the function.

PR #234, opened 2017-07-25
Removes state obsoleted by upstream patch in RHEL7/CentOS7

lorengordon:


lorengordon:
Modified PR to remove the sls and the call to it.

PR #236, opened 2017-07-28
Include Boot Arg

KevinPlus3:
Adding ability to include `boot=` argument to GRUB.

### DIFF
--- a/_modules/ash_linux.py
+++ b/_modules/ash_linux.py
@@ -42,6 +42,20 @@ def _move_boot_kernel(restore_bak):
     return old_path
 
 
+def _get_default_kernel():
+    """Obtain default kernel from grubby command."""
+    cmd = 'grubby --default-kernel'
+    return __salt__['cmd.run'](cmd, python_shell=True, output_loglevel='quiet')
+
+
+def _get_boot_mount():
+    """Obtain mount to path that kernel is located in."""
+    boot_path = _get_default_kernel()
+    boot_path = boot_path[:boot_path.rindex('/')] or '/'
+    cmd = 'findmnt -no source -T ' + boot_path
+    return __salt__['cmd.run'](cmd, python_shell=True, output_loglevel='quiet')
+
+
 def _modify_grub_file(rmv_fips_arg):
     """
     Insert/remove fips argument from GRUB file in /etc/default.
@@ -54,14 +68,27 @@ def _modify_grub_file(rmv_fips_arg):
     if rmv_fips_arg:
         result = __salt__['file.replace'](filepath, 'fips=1[ ]', '')
     else:
-        result = __salt__['file.search'](filepath, 'fips=1')
-        if not result:
+        grub_marker = 'GRUB_CMDLINE_LINUX="'
+        grub_args = []
+        check = __salt__['file.search'](filepath, 'boot=')
+        if not check:
+            # No boot= in grub, so find mount where kernel is located.
+            boot_mount = _get_boot_mount()
+            # Add boot= argument.
+            grub_args.append('boot={0} '.format(boot_mount))
+
+        check = __salt__['file.search'](filepath, 'fips=1')
+        if not check:
+            grub_args.append('fips=1 ')
+
+        if grub_args:
             result = __salt__['file.replace'](
-                filepath, 'GRUB_CMDLINE_LINUX="',
-                'GRUB_CMDLINE_LINUX="fips=1 '
+                filepath, grub_marker,
+                '{0}{1}'.format(grub_marker, ''.join(grub_args))
             )
         else:
             result = None
+
     return result
 
 
@@ -90,7 +117,7 @@ def _get_installed_dracutfips_pkgs():
 
 def _get_grub_args():
     """Obtain arguments line in grubby command."""
-    cmd = "grubby --info=ALL | grep args="
+    cmd = 'grubby --info=' + _get_default_kernel() + ' | grep args='
     return __salt__['cmd.run'](cmd, python_shell=True, output_loglevel='quiet')
 
 
@@ -226,10 +253,17 @@ def fips_enable():
             _move_boot_kernel(False)
             __salt__['cmd.run']("dracut -f", python_shell=False)
 
-        # Update grub.cfg file to add the fips agurment.
+        # Update grub.cfg file to add the fips and boot agurments.
+        grubby_args = []
         grub_args = _get_grub_args()
         if 'fips=1' not in grub_args:
-            cmd = 'grubby --update-kernel=ALL --args=fips=1'
+            grubby_args.append('fips=1')
+        if 'boot=' not in grub_args:
+            grubby_args.append('boot={0}'.format(_get_boot_mount()))
+        if grubby_args:
+            cmd = 'grubby --update-kernel=ALL --args="{0}"'.format(
+                ' '.join(grubby_args)
+            )
             __salt__['cmd.run'](cmd, python_shell=False)
             new['grubby'] = cmd
 


ferricoxide:
Haven't looked at the code, yet, but won't matter: can't be accepted till the checks stop failing. ;)

_modules/ash_linux.py
```diff
@@ -54,6 +54,29 @@ def _modify_grub_file(rmv_fips_arg):
...
+            # No boot= in grub, so find where kernel is located.
+            cmd = "grubby --default-kernel"
+            bt_path = __salt__['cmd.run'](
+                cmd, python_shell=True, output_loglevel='quiet'
+            )
+            bt_path = bt_path[:bt_path.rindex('/')]
+            if not bt_path:
+                bt_path = '/'
+            # Determine volume that is mounted to where kernel is located.
+            cmd = "df -h | grep " + bt_path + "$ | awk '{print $1}'"
```
ferricoxide: Oof... You can kill that grep. If you know your kernel file's path, you can just do `df  <flags> FILENAME` and it will return the directory it's in. For example:

~~~
$ df -Ph ~/dead.letter
Filesystem                    Size  Used Avail Use% Mounted on
/dev/mapper/OptionVG-dataVol   20G   15G  3.4G  82% /home
~~~

You should also be able to use `findmnt` (part of the `util-linux` RPM ...that's in the Core RPM-set):
~~~
$ findmnt -no source -T /boot/vmlinuz-3.10.0-514.26.2.el7.x86_64
/dev/sda
~~~

_modules/ash_linux.py
```diff
@@ -54,6 +54,29 @@ def _modify_grub_file(rmv_fips_arg):
     if rmv_fips_arg:
         result = __salt__['file.replace'](filepath, 'fips=1[ ]', '')
     else:
+        bt_result = __salt__['file.search'](filepath, 'boot=')
```
lorengordon: I have a very strong preference to spell out variables. My future self doesn't want to have to guess what `bt` means...

_modules/ash_linux.py
```diff
@@ -54,6 +54,29 @@ def _modify_grub_file(rmv_fips_arg):
...
+        if not bt_result:
+            # No boot= in grub, so find where kernel is located.
+            cmd = "grubby --default-kernel"
+            bt_path = __salt__['cmd.run'](
+                cmd, python_shell=True, output_loglevel='quiet'
+            )
+            bt_path = bt_path[:bt_path.rindex('/')]
```
lorengordon: How about `bt_path = bt_path[:bt_path.rindex('/')] or '/'` and then ditch the following `if` test?

_modules/ash_linux.py
```diff
@@ -62,7 +85,11 @@ def _modify_grub_file(rmv_fips_arg):
             )
         else:
             result = None
-    return result
+
+    if result:
+        return result
+    else:
```
lorengordon: Are we losing the `bt_result` diff or does `result` contain both diffs that this point? Perhaps we should construct the string to add considering both `boot=...` and `fips=1`, and call `file.replace()` only once?

_modules/ash_linux.py
```diff
@@ -54,6 +54,29 @@ def _modify_grub_file(rmv_fips_arg):
...
+            cmd = "df -h | grep " + bt_path + "$ | awk '{print $1}'"
+            bt_path = __salt__['cmd.run'](
+                cmd, python_shell=True, output_loglevel='quiet'
+            )
+            # Add boot= argument.
+            bt_result = __salt__['file.replace'](
+                filepath, 'GRUB_CMDLINE_LINUX="',
+                'GRUB_CMDLINE_LINUX="boot=' + bt_path + ' '
+            )
+        else:
```
lorengordon: Can use `bt_result = __salt__['file.search'](filepath, 'boot=') or None` above to set `bt_result` and ditch this `else` block.

_modules/ash_linux.py
```diff
@@ -54,14 +54,37 @@ def _modify_grub_file(rmv_fips_arg):
...
         result = __salt__['file.replace'](filepath, 'fips=1[ ]', '')
     else:
-        result = __salt__['file.search'](filepath, 'fips=1')
-        if not result:
+        args_to_add = 'GRUB_CMDLINE_LINUX="'
+        check = __salt__['file.search'](filepath, 'boot=')
+        if not check:
+            # No boot= in grub, so find where kernel is located.
+            cmd = "grubby --default-kernel"
+            bt_path = __salt__['cmd.run'](
```
lorengordon: Can we call this `boot_path` or whatever `bt` actually stands for, so my future self doesn't have to think about it to understand the variable name?

_modules/ash_linux.py
```diff
@@ -54,14 +54,37 @@ def _modify_grub_file(rmv_fips_arg):
     if rmv_fips_arg:
         result = __salt__['file.replace'](filepath, 'fips=1[ ]', '')
     else:
-        result = __salt__['file.search'](filepath, 'fips=1')
-        if not result:
+        args_to_add = 'GRUB_CMDLINE_LINUX="'
+        check = __salt__['file.search'](filepath, 'boot=')
+        if not check:
+            # No boot= in grub, so find where kernel is located.
+            cmd = "grubby --default-kernel"
```
lorengordon: I'd like to standardize on single quotes, rather than mixing and matching.

_modules/ash_linux.py
```diff
@@ -54,14 +54,37 @@ def _modify_grub_file(rmv_fips_arg):
...
+            bt_path = __salt__['cmd.run'](
+                cmd, python_shell=True, output_loglevel='quiet'
+            )
+            bt_path = bt_path[:bt_path.rindex('/')] or '/'
+
+            # Determine volume that is mounted to where kernel is located.
+            cmd = "findmnt -no source -T " + bt_path
```
lorengordon: Single quotes

_modules/ash_linux.py
```diff
@@ -54,14 +54,37 @@ def _modify_grub_file(rmv_fips_arg):
...
+            bt_path = __salt__['cmd.run'](
+                cmd, python_shell=True, output_loglevel='quiet'
+            )
+
+            # Add boot= argument.
+            args_to_add += "boot=" + bt_path + ' '
```
lorengordon: Single quotes

_modules/ash_linux.py
```diff
@@ -54,14 +54,37 @@ def _modify_grub_file(rmv_fips_arg):
...
+
+        check = __salt__['file.search'](filepath, 'fips=1')
+        if not check:
+            args_to_add += "fips=1 "
```
lorengordon: single quotes

_modules/ash_linux.py
KevinPlus3: :thumbsup:

_modules/ash_linux.py
KevinPlus3: 👍

_modules/ash_linux.py
```diff
@@ -54,14 +54,37 @@ def _modify_grub_file(rmv_fips_arg):
...
+
+        if args_to_add.endswith(' '):
```
lorengordon: I see why you're doing this, but it feels a little magical. How about defining it like this way up at the beginning of this else block:

```
grub_marker = 'GRUB_CMDLINE_LINUX="'
grub_args = []
```

Then do the checks and add items to `grub_args` as needed, modify this `if` block to be:

```
if grub_args:
    result = __salt__['file.replace'](filepath, grub_marker, '{0} '.format(' '.join(grub_args)))
```

KevinPlus3: That is ssseeeexy. Consider it done.

lorengordon: Actually the repl is probably easier, because you're adding the space to each argument:
```
if grub_args:
    result = __salt__['file.replace'](filepath, grub_marker, ''.join(grub_args))
```

lorengordon: Oh wait, need to prepend `grub_marker` on the repl...
```
if grub_args:
    result = __salt__['file.replace'](filepath, grub_marker, '{0}{1}'.format(grub_marker, ''.join(grub_args)))
```

PR #238, opened 2017-08-07
Had to sledghammer it:

ferricoxide:
* Moving the ipv4 forwarding stuff to a 'last-executed' filename in the
  /etc/sysctl.d directory should have been more than enough
* However, it wasn't, so had to add a "make sure these conflicting,
  Samsonite-gorilla RPMs aren't present to cause problems" package-
  removal operation to ensure The Right Thing(TM) happens with older
  AMIs.

### DIFF
--- a/ash-linux/el6/STIGbyID/cat2/V38511.sls
+++ b/ash-linux/el6/STIGbyID/cat2/V38511.sls
@@ -17,7 +17,7 @@
 
 {%- set stigId = 'V38511' %}
 {%- set helperLoc = 'ash-linux/el6/STIGbyID/cat2/files' %}
-{%- set chkFile = '/etc/sysctl.conf' %}
+{%- set chkFile = '/etc/sysctl.d/99-ip_forward.conf' %}
 {%- set parmName = 'net.ipv4.ip_forward' %}
 {%- set parmVal = '0' %}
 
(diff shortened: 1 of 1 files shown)


ash-linux/el6/STIGbyID/cat2/V38511.sls
```diff
@@ -27,18 +27,27 @@ script_{{ stigId }}-describe:
...
+  file.managed:
     - name: '{{ chkFile }}'
-    - pattern: '^{{ parmName }} =.*$'
-    - repl: |
-        # Added {{ parmName }} define per STIG-ID: {{ stigId }}
-        {{ parmName }} = {{ parmVal }}
-    - unless: 'grep -Ew "{{ parmName }}" {{ chkFile }} && \
-               grep -Ew "{{ parmName }}" {{ chkFile }} | \
-               grep -E "{{ parmVal }}$"'
+    - contents: |
```
lorengordon: Do we need to worry whether the `/etc/sysctl.d/` directory exists? If so, add this option to the state:
```
    - makedirs: True
```

ferricoxide: Netgative: it's part of the `initscripts` RPM. Minus that RPM, EL6 no-workie. If someone is going around randomly deleting critical directories, I'm disinclined to provide them kevlar loafers.

PR #241, opened 2017-08-09
Safety implemented via Salt's file.managed method

ferricoxide:
This method:
* Creates a file that would normally be created via the `firewalld-offline-cmd` command-line utility
* Issues a firewalld service-restart if the target file is new or updated

Was going to code up other methods, but this is both Salt-ier and less prone to shortcomings in Saltstack's current implementation of firewalld.

Note: this module isn't currently linked into any of the `init.sls`, `top.sls` or equivalent activation method.


lorengordon:
Tested? Or need me to run it through?

ferricoxide:
Method tested before push but have not tested as part of a watchmaker run (see PR's comment about not being linked to an activating-method)

PR #243, opened 2017-09-28
Change from file to cmd method

ferricoxide:
This is _a_ fix method. Validated that, even invoking `reload`, it works. Primary benefit of this method is it's additive in nature vice the previous method's absolutist posture:
- rules are added to existing `direct` rules, if such exist, rather than replacing existing rule-set
- if exact rules already exist, no changes to config are made
- if other rules already exist but don't overlap (e.g. a tenant has their own direct rules or we ever need to add more direct rules), these will simply interleave themselves with any existing rules.

Might want to bundle this up into a sourced script to ensure that the exit returned isn't simply the exit of the final called-command.


ferricoxide:
The changes pushed today result in the requisite rules getting created with notionally more-meaningful outputs produced (and with the capability of running with more-verose outputs as necessary).  A system remediated via this PR and with default-zone changed to `drop` retains expected connectivity

PR #249, opened 2018-08-06
Ensures firewalld is installed, and the service is running and enabled

lorengordon:


lorengordon:
Travis failure is unrelated. That will be addressed separately.

PR #251, opened 2018-08-07
Ensures /etc/fstab exists in the docker container

lorengordon:


lorengordon:
Well this works, not sure it's the "right" way, but... 🤷‍♂️

ferricoxide:
> Well this works, not sure it's the "right" way, but... 🤷‍♂️ 

lulz

lorengordon:
> use the docker command to create something that will prevent the stat from failing on /etc/fstab

that's it, yep. couldn't figure out what the problem was, so basically just bailing for now. other option i suppose could be to update the salt states to add `ignore_if_missing=True` to the file.search command...

ferricoxide:
I'm thinking we probably don't want to potentially (and, likely, "silently") break the actual _running_ of the formula for the sake of making a test pass (by adding that `ignore...`)?

lorengordon:
My thinking also, hence I think the current workaround is better... :D

PR #252, opened 2018-08-07
Pins docker image versions so dependabot can update them

lorengordon:


lorengordon:
While, this works, we can leave this one open a while, as dependabot may come through with a better way in response to an issue I opened on their tracker.

lorengordon:
@plus3it/watchmaker this is ready for review

lorengordon:
Oops, meant @plus3it/salt-formulas...

PR #259, opened 2019-01-28
Minor clean-up and addresses several issues

### LINKED ISSUE #254: IPv6 Disablement Breaks Default Postfix Config
SCAP/STIG-related disablement of IPv6 causes default Postfix configuration to become non-functional. Once hardening is applied, all postfix-related operations will fail with error text similar to:

~~~~
<POSTFIX_ELEMENT>: fatal: parameter inet_interfaces: no local interface found for ::1
~~~~

And system logs will contain:

~~~~
Oct 02 13:22:35 master-test systemd[1]: Starting Postfix Mail Transport Agent...
Oct 02 13:22:35 master-test aliasesdb[1519]: /usr/sbin/postconf: fatal: parameter inet_interfaces: no local interface found for ::1
Oct 02 13:22:36 master-test postfix/sendmail[1771]: fatal: parameter inet_interfaces: no local interface found for ::1
Oct 02 13:22:36 master-test aliasesdb[1519]: newaliases: fatal: parameter inet_interfaces: no local interface found for ::1
Oct 02 13:22:36 master-test postfix[1792]: fatal: parameter inet_interfaces: no local interface found for ::1
~~~~

Need to add a salt-state to comment out (or wholly-delete) the:

~~~~
::1     localhost       localhost.localdomain   localhost6      localhost6.localdomain6
~~~~
 
Line from the `/etc/hosts` file or change the value of:

~~~~
inet_protocols = all
~~~~

to:

~~~~
inet_protocols = ipv4
~~~~

In `/etc/postfix/main.cf`

### LINKED ISSUE #258: Watchmaker generates error when ran a second time on EL6 system
When Watchmaker is applied a second time on an EL6 system, the following error appears in `/var/log/watchmaker/watchmaker.log`:

```
[watchmaker.managers.base.SaltLinux][ERROR][1560]: Command stderr: b"[ERROR   ] Command '[u'sysctl', u'-n', u'net.ipv6.conf.all.disable_ipv6']' failed with return code: 255"
[watchmaker.managers.base.SaltLinux][ERROR][1560]: Command stderr: b'[ERROR   ] stdout: error: "net.ipv6.conf.all.disable_ipv6" is an unknown key'
```

It appears, in EL6, the disablement of IPv6 causes all IPv6 kernel parameters to be removed and the use of Salt `sysctl.get` generates an error when attempting to access these IPv6 parameters.

eemperor:
This pull request addresses the following:
- Fixes issues caused by disablement of `ipv6`
  - Fixes #254
  - Fixes #258 
- Updates usage of Salt to address depricated functions.

### DIFF
--- a/_modules/ash_linux.py
+++ b/_modules/ash_linux.py
@@ -8,7 +8,10 @@ import os
 import re
 import spwd
 
-import salt.utils
+try:
+    from salt.utils.files import fopen
+except ImportError:
+    from salt.utils import fopen
 
 __virtualname__ = 'ash'
 
@@ -303,7 +306,7 @@ def fips_status():
         salt '*' ash.fips_status
     """
     try:
-        with salt.utils.fopen('/proc/sys/crypto/fips_enabled', 'r') as fle:
+        with fopen('/proc/sys/crypto/fips_enabled', 'r') as fle:
             return 'enabled' if fle.read().strip() == '1' else 'disabled'
     except (IOError, FileNotFoundError):
         return 'disabled'
--- a/ash-linux/el6/STIGbyID/cat2/V38444.sls
+++ b/ash-linux/el6/STIGbyID/cat2/V38444.sls
@@ -3,11 +3,11 @@
 # Version:	RHEL-06-000523
 # Finding Level:	Medium
 #
-#     In "ip6tables" the default policy is applied only after all the 
-#     applicable rules in the table are examined for a match. Setting 
-#     the default policy to "DROP" implements proper design for a 
-#     firewall, i.e., any packets which are not explicitly permitted 
-#     should not be accepted. 
+#     In "ip6tables" the default policy is applied only after all the
+#     applicable rules in the table are examined for a match. Setting
+#     the default policy to "DROP" implements proper design for a
+#     firewall, i.e., any packets which are not explicitly permitted
+#     should not be accepted.
 #
 #  CCI: CCI-000066
 #  NIST SP 800-53 :: AC-17 e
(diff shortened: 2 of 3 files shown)


ferricoxide:
I'd be concerned about leveraging simple file existence-checks on anything in `/proc`. Some subsystems' `/proc` handle will continue to exist &mash; but with a different value in a disabled-state than in an enabled state &mdash; whether a given subsystem is active or not (particularly in a changed-state). I dunno that this is true of IPv6.

lorengordon:
> I'd be concerned about leveraging simple file existence-checks on anything in `/proc`

I'm not sure there's much of an alternative. Salt is using `sysctl -n net.ipv6.conf.all.disable_ipv6` as it should, but sysctl bombs when the file in `/proc` does not exist.

We could perhaps use `cmd.run` to execute `systcl -n` directly, and test the return code _or_ the return value?

eemperor:
@ferricoxide EL7 behaves as you describe.  However, my update is isolated to only EL6.  If you still have concerns, I can try the `cmd.run` approach as @lorengordon suggests.
 
@lorengordon You beat me to the punch.  I was thinking along the same lines.

ferricoxide:
Sorry. Didn't mean to imply the specific use-case was bad, simply saying that it can _generically_ be a point of concern. That said, assuming this has been tested to do what it's intended to (and verified to work), then I've no objections to the method.

eemperor:
I verified it's working on the last CentOS6 spel AMI.  I'll check on RedHat6 as well.

ferricoxide:
Cool. Then, I've no objections. =)

eemperor:
👍

PR #262, opened 2019-02-14
Add logic to set GRUB2 password

### LINKED ISSUE #260: Set GRUB2 Password
See [watchmaker issue 704-C](https://github.com/plus3it/watchmaker/issues/704#issuecomment-432353974)

ferricoxide:
Closes #260 

Use logic to _always_ create/overwrite the `/boot/grub2/user.cfg` file if Pillar value is set while only creating if Pillar value is undef/null _and_ `/boot/grub2/user.cfg` file doesn't already exist


ferricoxide:
Careful, [there](https://www.dailydot.com/unclick/69-nice-meme-twitter/)...

PR #263, opened 2019-02-14
Add state to apply audit rules changes without a reboot

### LINKED ISSUE #257: Load `/etc/audit/audit.rules` without rebooting
Currently `/etc/audit/audit.rules` is not loaded by default and instead relies on a reboot. When combined with the `--no-reboot` switch in Watchmaker the rules are never applied.

Please include `augenrules --load` after configuring the rules.

eemperor:
Fixes #257


eemperor:
Recommended changes made.

PR #267, opened 2019-02-20
Add handlers for CIS 5.2.3 that some scanners/profiles check for

### LINKED ISSUE #253: Add formula to ensure "LogLevel" is set to "VERBOSE" in sshd_config
Standard `LogLevel` is `INFO`. This log-level does not capture connecting users's SSH keys. Probably want to change this &mdash; at least on hosts that are not AD-integrated.

ferricoxide:
Probably need to open a BugZilla to get this removed from the C2S profile. At any rate, closes #253

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -66,6 +66,11 @@
 ##  Set a localized/custom password on the GRUB boot-loader
 ##     grub-passwd: <POLICY_COMPLIANT_PASSWORD_STRING>
 
+##  Ensure that the LogLevel parm in sshd_config is set to the
+##  site-mandated value
+##     sshd-loglevel:  <QUIET|FATAL|ERROR|INFO|VERBOSE|DEBUG|DEBUG1|DEBUG2|DEBUG3>
+##  (See man page for valid values and associated output/warnings)
+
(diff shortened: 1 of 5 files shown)


ferricoxide:
Note: opened [Bugzilla 1679253](https://bugzilla.redhat.com/show_bug.cgi?id=1679253) to see if vendor will remove this superfluous test from the C2S profile.

pillar.example
```diff
@@ -66,6 +66,10 @@
 ##  Set a localized/custom password on the GRUB boot-loader
 ##     grub-passwd: <POLICY_COMPLIANT_PASSWORD_STRING>
 
+##  Ensure that the LogLevel parm in sshd_config is set to the
+##  site-mandated value
+##     sshd-loglevel:  Info
```
lorengordon: is this case-sensitive in the ssdh_config? you have `Info` here and `INFO` as the default in the state...

ferricoxide: Yeah, it's case-sensitive. So, the example value is wrong. Shoulda just put a reference to the man page.

PR #271, opened 2019-02-26
Ensure a `MaxAuthTries` value is explicitly set 

### LINKED ISSUE #270: Add Content to Ensure `MaxAuthTries` Value is set in SSH Daemon's Configuration
See [upstream issue](https://github.com/plus3it/watchmaker/issues/801)

ferricoxide:
Closes #270 

Note: for whatever reason, the `OS_VERSION=6 SALT_STATE=ash-linux.scap` is refusing to go green for me. It's not relevant to this change, since it's a change wholly within the EL7 tree, but... May require a force-merge if Travis is going to continue to flake out on it, today.

### DIFF
--- a/ash-linux/el6/SCAPonly/low/CCE-26340-0.sls
+++ b/ash-linux/el6/SCAPonly/low/CCE-26340-0.sls
@@ -26,7 +26,6 @@ script_{{ scapId }}-describe:
 append_{{ scapId }}-directive:
   file.append:
     - name: '{{ moduleConf }}'
-    - mode: '0644'
     - text: |
         # Added per SCAP-ID {{ scapId }}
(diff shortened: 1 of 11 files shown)


lorengordon:
The travis error is actually a valid error. Salt 2019.2.0 was released yesterday, and `mode` is no longer a valid argument to `file.append`. It just happens that this is the only state where we are using `file.append` (rather than `file.replace`).

I'd say you can fix it here, or fix it in another pull request and then rebase this one once that is merged. Or pin the test to install salt 2018.3.x.

ferricoxide:
Goddammit...

ferricoxide:
Really shitty part is those were all in the "CCE" handlers (not even the EL6 "STIG" handlers).

PR #289, opened 2020-05-18
Add retry to package install and start

YakDriver:


ash-linux/el6/STIGbyID/cat3/V38669.sls
```diff
@@ -40,10 +40,20 @@ notify_{{ stigId }}-postfix:
...
 svc_{{ stigId }}-postfixEnabled:
   service.enabled:
     - name: '{{ wantedPkg }}'
+    - retry:
+        attempts: 20
+        until: True
+        interval: 2
+        splay: 10
 
 svc_{{ stigId }}-postfixRunning:
```
lorengordon, line 58: So, all those retries are on states other than the one that was failing?

YakDriver: Ergh... I need to double check... I posted it was on running but I remember it being on starting so one of them is wrong.

YakDriver: You are right. Running was failing.

PR #300, opened 2021-02-24
Make more reliable in face of misconfigured yum repositories

### LINKED ISSUE #299: [BUG] RHEL-07-040560 State Fails If Yum Repository Is Missing Groups Metadata
**Describe the bug**

When the `ash-linux/el7/STIGbyID/cat2/RHEL-07-040560.sls` is run on an EL7 host with a misconfigured yum repository (the repository is missing groups metadata), it ABENDs

**Severity**

- [x] Completely Broken (No work-around evident)
- [ ] Moderately Broken (Trivial work-around)

**To Reproduce**
1. Launch a VM configured to use a private EL7 yum repository
2. Invoke SaltStack formula that references the `ash-linux.el7.STIGbyID.cat2.RHEL-07-040560` state
3. State fails with:
~~~
Salt state execution failed:
- "Rendering SLS 'base:ash-linux.el7.STIGbyID.cat2.RHEL-07-040560' failed: Problem\
    \ running salt function in Jinja template: Group 'X Window System' not found;\
    \ line 17\n\n---\n[...]\n#    NIST SP 800-53 Revision 4 :: CM-6 b \n#\n#################################################################\n\
    {%- set stig_id = 'RHEL-07-040560' %}\n{%- set helperLoc = 'ash-linux/el7/STIGbyID/cat2/files'\
    \ %}\n{%- set rpmGrpLst = salt.pkg.group_info('X Window System') %}    <======================\n\
    {%- set targRpm = \"xorg-x11-server-common\" %}\n{%- set skipIt = salt.pillar.get('ash-linux:lookup:skip-stigs',\
    \ []) %}\n\nscript_{{ stig_id }}-describe:\n  cmd.script:\n[...]\n---"
~~~

**Expected behavior**

State should run without error.

**Deviance Description**

State errors out, even if excluded in the ash-linux-forumla pillar's `linux:lookup:skip-stig` list-var.

**Screenshots**

**Additional context**

State only fails when environment is configured to use a (private) yum repository that lacks RPM-groups metadata (misconfigured repository):
~~~
# yum group info 'X Window System'
Loaded plugins: fastestmirror
There is no installed groups file.
Maybe run: yum groups mark convert (see man yum)
Loading mirror speeds from cached hostfile
Warning: group/environment X Window System does not exist.
~~~

**Fix Suggestions**
Update the:
~~~
{%- set rpmGrpLst = salt.pkg.group_info('X Window System') %}
~~~
[Line](https://github.com/plus3it/ash-linux-formula/blob/master/ash-linux/el7/STIGbyID/cat2/RHEL-07-040560.sls#L17) to tolerate a null-return.

ferricoxide:
Closes #299

### DIFF
--- a/ash-linux/el7/STIGbyID/cat2/RHEL-07-040560.sls
+++ b/ash-linux/el7/STIGbyID/cat2/RHEL-07-040560.sls
@@ -1,28 +1,26 @@
-# Finding ID:	RHEL-07-040560
-# Version:	RHEL-07-040560_rule
-# SRG ID:	SRG-OS-000480-GPOS-00227
-# Finding Level:	medium
-# 
-# Rule Summary:
-#	An X Windows display manager must not be installed unless approved.
+# Finding ID:   RHEL-07-040560
+# Version:      RHEL-07-040560_rule
+# SRG ID:       SRG-OS-000480-GPOS-00227
+# Finding Level:        medium
 #
-# CCI-000366 
-#    NIST SP 800-53 :: CM-6 b 
-#    NIST SP 800-53A :: CM-6.1 (iv) 
-#    NIST SP 800-53 Revision 4 :: CM-6 b 
+# Rule Summary:
+#       An X Windows display manager must not be installed unless approved.
+#
+# CCI-000366
+#    NIST SP 800-53 :: CM-6 b
+#    NIST SP 800-53A :: CM-6.1 (iv)
+#    NIST SP 800-53 Revision 4 :: CM-6 b
 #
 #################################################################
 {%- set stig_id = 'RHEL-07-040560' %}
 {%- set helperLoc = 'ash-linux/el7/STIGbyID/cat2/files' %}
-{%- set rpmGrpLst = salt.pkg.group_info('X Window System') %}
+{%- set hazGroupMeta = salt.pkg.group_list()['available'] %}
 {%- set targRpm = "xorg-x11-server-common" %}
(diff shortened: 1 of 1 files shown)


ash-linux/el7/STIGbyID/cat2/RHEL-07-040560.sls
```diff
@@ -34,7 +32,10 @@ purge_{{ stig_id }}-{{ targRpm }}:
   pkg.removed:
     - pkgs:
       - '{{ targRpm }}'
-  {%- for rpm in rpmGrpLst %}
+  {%- if hazGroupMeta and 'X Window System' in hazGroupMeta %}
+    {%- set rpmGrpLst = salt.pkg.group_info('X Window System') %}
+    {%- for rpm in rpmGrpLst %}
```
lorengordon, line 37: i know i had it like this in my example, but we don't really need to set the var, since we don't re-use it...

```suggestion
    {%- for rpm in salt.pkg.group_info('X Window System') %}
```

lorengordon (approved):
Approving with one nit/suggestion. Will re-approve if you choose to change it...

ash-linux/el7/STIGbyID/cat2/RHEL-07-040560.sls
ferricoxide: At first, I was like, "but then you're calling `salt.pkg.group_info` twice". Sometimes, GitHub's presentation of code-comments don't help with clarity.

lorengordon (approved):

PR #308, opened 2021-03-11
Gracefully handle lack of an /etc/sysctl.conf file

### LINKED ISSUE #307: [BUG] Eliminate assumption of existence of /etc/sysctl.conf
**Describe the bug**

On systems that don't actually have an `/etc/sysctl.conf` file, the `el7/STIGbyId/cat2/RHEL-07-040860.sls` states will fail

**Severity**

- [X] Moderately Broken (Trivial work-around)

**Expected behavior**

The `el7/STIGbyId/cat2/RHEL-07-040860.sls` states will succeed even if the target system is missing the `/etc/sysctl.conf` file.

**Fix Suggestions**

PR inbound...

ferricoxide:
Closes #307

### DIFF
--- a/ash-linux/el7/STIGbyID/cat2/RHEL-07-040860.sls
+++ b/ash-linux/el7/STIGbyID/cat2/RHEL-07-040860.sls
@@ -44,3 +44,5 @@ file_{{ stig_id }}-{{ parmName }}:
     - not_found_content: |-
         # Inserted per STIG {{ stig_id }}
         {{ parmName }} = {{ parmValuTarg }}
+    - require:
+      - file: 'file-exists_{{ stig_id }}-{{ cfgFile }}'
--- a/ash-linux/el7/STIGbyID/cat2/RHEL-07-040860.sls
+++ b/ash-linux/el7/STIGbyID/cat2/RHEL-07-040860.sls
@@ -2,14 +2,14 @@
 # Version:	RHEL-07-040860_rule
 # SRG ID:	SRG-OS-000480-GPOS-00227
 # Finding Level:	medium
-# 
+#
 # Rule Summary:
 #	The system must not forward IPv6 source-routed packets.
 #
-# CCI-000366 
-#    NIST SP 800-53 :: CM-6 b 
-#    NIST SP 800-53A :: CM-6.1 (iv) 
-#    NIST SP 800-53 Revision 4 :: CM-6 b 
+# CCI-000366
+#    NIST SP 800-53 :: CM-6 b
+#    NIST SP 800-53A :: CM-6.1 (iv)
+#    NIST SP 800-53 Revision 4 :: CM-6 b
 #
 #################################################################
 {%- set stig_id = 'RHEL-07-040860' %}
@@ -26,8 +26,14 @@ script_{{ stig_id }}-describe:
 
 sysctl_{{ stig_id }}-{{ parmName }}:
   sysctl.present:
-    - name: '{{ parmName }}' 
+    - name: '{{ parmName }}'
     - value: '{{ parmValuTarg }}'
+
+file-exists_{{ stig_id }}-{{ cfgFile }}:
+  file.touch:
+    - name: '{{ cfgFile }}'
+    - unless:
+      - 'test -e {{ cfgFile }}'
 
 file_{{ stig_id }}-{{ parmName }}:
   file.replace:


ash-linux/el7/STIGbyID/cat2/RHEL-07-040860.sls
```diff
@@ -26,9 +26,15 @@ script_{{ stig_id }}-describe:
 
 sysctl_{{ stig_id }}-{{ parmName }}:
   sysctl.present:
-    - name: '{{ parmName }}' 
+    - name: '{{ parmName }}'
     - value: '{{ parmValuTarg }}'
```
eemperor, line 31: Should add the `require` option to control the ordering.

```suggestion
    - require:
      - file: file-exists_{{ stig_id }}-{{ cfgFile }}
```

ash-linux/el7/STIGbyID/cat2/RHEL-07-040860.sls
```diff
@@ -26,9 +26,15 @@ script_{{ stig_id }}-describe:
...
 
+file-exists_{{ stig_id }}-{{ cfgFile }}:
+  file.touch:
+    - name: '{{ cfgFile }}'
+    - unless:
+      - 'test -e {{ cfgFile }}'
+
 file_{{ stig_id }}-{{ parmName }}:
   file.replace:
     - name: '{{ cfgFile }}'
```
eemperor, line 40: Same here.
  
```suggestion
    - name: '{{ cfgFile }}'
    - require:
      - file: file-exists_{{ stig_id }}-{{ cfgFile }}
```

I want to suggest adding it after `- not_found_content:` but GitHub is not letting me for some reason...

ferricoxide: Web interface makes it not clear what you're suggesting?

ferricoxide: Ah... Ok, think I see what you're saying. Basically, while a given state-file's stanzas tend to be processed serially, you're saying "put a `require` from `file_{{ stig_id }}-{{ parmName }}` to `file-exists_{{ stig_id }}-{{ cfgFile }}` to forcibly-establish an ordered-linkage"?

eemperor: Yup.  Thats it.

ash-linux/el7/STIGbyID/cat2/RHEL-07-040860.sls
ferricoxide: There's no need for a linkage of the in-memory content (which `sysctl.present` stanza establishes) and the config-file content (that the `file.replace` establishes. There _shouldn't_ be any need for a `requires` since, at no point, is the state-file doing the equivalent of either a `sysctl reload`/`sysctl --system` or a `sysctl -p <cfgFile>`.

ash-linux/el7/STIGbyID/cat2/RHEL-07-040860.sls
ferricoxide: K. Incoming.

ash-linux/el7/STIGbyID/cat2/RHEL-07-040860.sls
eemperor: ok

eemperor (approved):
👍

ash-linux/el7/STIGbyID/cat2/RHEL-07-040860.sls
```diff
@@ -26,9 +26,15 @@ script_{{ stig_id }}-describe:
 
 sysctl_{{ stig_id }}-{{ parmName }}:
   sysctl.present:
-    - name: '{{ parmName }}' 
+    - name: '{{ parmName }}'
     - value: '{{ parmValuTarg }}'
 
+file-exists_{{ stig_id }}-{{ cfgFile }}:
+  file.touch:
```
lorengordon, line 33: fwiw, while this works, salt recommends using `file.managed` when just wanting to ensure that a file exists...

* https://docs.saltproject.io/en/latest/ref/states/all/salt.states.file.html#salt.states.file.touch

ferricoxide: Ah. Always thought `file.managed` was for installing files - either from a source or with a `content:` string ...though I guess using `file.managed` also allows you to set owner/mode/sel-label while `file.touch` assumes that the state's run-as user will have an appropriate umask?

PR #310, opened 2021-03-17
Ensure no (competing) attempts to set TMOUT

### LINKED ISSUE #304: [BUG] cat2/RHEL-07-040160.sls needs to account for already-present TMOUT settings
**Describe the bug**

If a to-be-hardened host already has `TMOUT` set in `/etc/profile.d/*`, after the named-state runs, interactive users will receive errors about the `TMOUT` variable being immutable. This can cause other automations to fail.  

**Severity**

- [ ] Moderately Broken (Trivial work-around)
- [X] Nuisance (Functions but untrapped errors can slip through)

**To Reproduce**
1. Launch an instance with the bash-shell's `TMOUT` variable already set via an `/etc/profile.d` file
2. Run watchmaker
3. Logout of system
4. Log back into hardened system
5. Error message printed that `TMOUT` may not be overridden

**Expected behavior**

`TMOUT` set only once across `/etc/profile` or `/etc/profile.d/*` files: interactive logins experience no errors related to `TMOUT` being set read-only

**Fix Suggestions**

Have state check for presence of `TMOUT` sets in other `/etc/profile.d/*` files and comment them out

ferricoxide:
Closes #304 

Will look for any competing values of:

- `TMOUT=`
- `readonly TMOUT=`
- `declare -r TMOUT=`

Set in _any_ of the `/etc/profile.d` files

### DIFF
--- a/ash-linux/el7/STIGbyID/cat2/RHEL-07-040160.sls
+++ b/ash-linux/el7/STIGbyID/cat2/RHEL-07-040160.sls
@@ -45,5 +45,5 @@ file_{{ stig_id }}-{{ cfgExtra }}:
     - backup: False
     - name: '{{ cfgExtra }}'
     - pattern: '^([\s]*)(|readonly|declare -r)([\s]*TMOUT=.*)$'
-    - repl: '# \1\2\3'
+    - repl: '#\1\2\3'
 {%- endfor %}
--- a/ash-linux/el7/STIGbyID/cat2/RHEL-07-040160.sls
+++ b/ash-linux/el7/STIGbyID/cat2/RHEL-07-040160.sls
@@ -2,7 +2,7 @@
 # Version:	RHEL-07-040160_rule
 # SRG ID:	SRG-OS-000163-GPOS-00072
 # Finding Level:	medium
-# 
+#
 # Rule Summary:
 #	All network connections associated with a communication
 #	session must be terminated at the end of the session or after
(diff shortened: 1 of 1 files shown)


ash-linux/el7/STIGbyID/cat2/RHEL-07-040160.sls
```diff
@@ -38,4 +38,12 @@ file_{{ stig_id }}-{{ cfgFile }}:
...
         readonly {{ parmName }}={{ parmValu }}
- 
+
+{%- for cfgExtra in salt.file.find('/etc/profile.d', maxdepth='0', type='f') %}
+file_{{ stig_id }}-{{ cfgExtra }}:
+  file.replace:
+    - backup: False
+    - name: '{{ cfgExtra }}'
+    - pattern: '^([\s]*)(|readonly|declare -r)([\s]*TMOUT=.*)$'
+    - repl: '# \1\2\3'
```
lorengordon, line 48: Yeah, super nitpicky, but I'd suggest _not_ injecting the space here. The pattern is preserving leading spaces already. So when _uncommenting_, we kind of get into this situation where the user needs to know, is space one I should preserve or not. But not blocking if you feel more strongly about keeping it.

```suggestion
    - repl: '#\1\2\3'
```

lorengordon (approved):
Approving with one non-blocking suggestion....

lorengordon (approved):
👍

PR #358, opened 2023-04-11
Conditionally Add `fapolicyd` Exceptions for AWS CLIv2

### LINKED ISSUE #357: [Feature Request] Need `fapolicyd`  exception for AWS CLI

With later versions of the RHEL8 STIGs, the use of application-whitelisting by way of the `fapolicyd`  service is required. Due to the method that the AWS CLI is delivered by Amazon, the `fapolicyd`  service prevents its execution. Need to add an `fapolicyd` exception for the AWS CLI.

Add a rule-file to `/etc/fapolicyd/rules.d/` that allows AWS CLI to function. A file with contents similar to:

~~~
allow perm=any all : dir=/usr/local/aws-cli/v2/ type=application/x-sharedlib trust 1
allow perm=any all : dir=/usr/local/aws-cli/v2/ type=application/x-executable trust 1
~~~

Will be sufficient to allow the AWS CLI to function _and_ continue to function after upgrades

AWS's initial recommendation was to create the exception using `fapolicyd-cli --file add /usr/local/bin/aws`; however, the resultant exception is too strict. The exception is for the resolved-path to the AWS CLI and requires the resolved-path's size _and_ checksum to never change. Using the AWS-suggested method would mean that any upgrades to the AWS CLI would also require an update to the exception.

**Additional context**

ferricoxide:
Closes #357


ash-linux/el8/VendorSTIG/remediate.sls
```diff
@@ -21,6 +21,7 @@
 ) %}
 {%- set pillProf = salt.pillar.get('ash-linux:lookup:scap-profile', 'common') %}
 {%- set scapProf = 'xccdf_org.ssgproject.content_profile_' ~ pillProf %}
+{%- set awscli_filetypes = [ 'executable', 'sharedlib' ] %}
```
eemperor, line 24: Why are you setting this here also?

ash-linux/el8/VendorSTIG/aws_cli_v2.sls
```diff
@@ -0,0 +1,41 @@
...
+    - repl: 'allow perm=any all : dir=/usr/local/aws-cli/v2/ type=application/x-{{ fileType }} trust 1'
+    - require:
+      - file: 'Ensure fapolicyd exception-file exists'
+{%- endfor %}
+
+Reload fapolicyd config:
+  cmd.run:
+    - name: '/usr/sbin/fapolicyd-cli -u'
+    - cwd: '/root'
+    - stateful: False
```
eemperor, line 41: Probably needs a `- require:` line to ensure the reload runs after the previous states and maybe a `- onlyif` also to check if it's even necessary.

ash-linux/el8/VendorSTIG/remediate.sls
ferricoxide: leftovers from before I split the files out.

ash-linux/el8/VendorSTIG/aws_cli_v2.sls
ferricoxide: That's what the

~~~
    - onchanges_in:
      - cmd: 'Reload fapolicyd config'
~~~

in the `Exempt AWS CLI v2 From fapolicyd ({{ fileType }}):` state does.

eemperor (approved):

PR #378, opened 2023-04-19
Addresses V-230364: Set each user's minimum password-age policy-value

### LINKED ISSUE #367: [Feature Request] The `root` User Must Not Be Allowed To Change Its Password More Than Once Per 24 Hours

Per V-230364/RHEL-08-020180, `root` have a minimum password age of `0` or blank.  Check-condition of:

~~~
awk -F: '$4 < 1 {print $1 " " $4}' /etc/shadow
~~~

Must not return for `root`.

Ensure that fourth `/etc/shadow` field for `root`  user is `1` or greater 

**Additional context**

On an EC2 launched from the spel or Red Hat official AMIs that has had no hardening applied, executing the check-condition results in output similar to:

~~~
root 0
bin 0
daemon 0
adm 0
lp 0
sync 0
shutdown 0
halt 0
mail 0
operator 0
games 0
ftp 0
nobody 0
dbus
tss
systemd-coredump
systemd-resolve
unbound
polkitd
sssd
chrony
sshd
maintuser 0
ssm-user 0
~~~

Executing the check-condition after `oscap --remediate` runs, output is similar to:

~~~
fapolicyd
postfix
~~~

Need a further module that ensures that users created after `oscap --remediate` runs _also_ have their `/etc/shadow` min-days field set to 1 or greater.

ferricoxide:
Sets minimum password age of "1" for every enumerated user that has either a null or sub-`1` value for field number 4 in /etc/shadow.

Closes #367

### DIFF
--- a/ash-linux/el8/STIGbyID/cat2/RHEL-08-020180.sls
+++ b/ash-linux/el8/STIGbyID/cat2/RHEL-08-020180.sls
@@ -1,0 +1,46 @@
+# Ref Doc:    STIG - RHEL 8 v1r9
+# Finding ID: V-230364
+# Rule ID:    SV-230364r627750_rule
+# STIG ID:    RHEL-08-020180
+# SRG ID:     SRG-OS-000075-GPOS-00043
+#
+# Finding Level: medium
+#
+# Rule Summary:
+#       Passwords managed by the operating system  must have a 24 hours/1
+#       day minimum password lifetime restriction in /etc/shadow.
+#
+# References:
+#   CCI:
+#     - CCI-000198
+#   NIST SP 800-53 :: IA-5 (1) (d)
+#   NIST SP 800-53A :: IA-5 (1).1 (v)
+#   NIST SP 800-53 Revision 4 :: IA-5 (1) (d)
+#
+###########################################################################
+{%- set stig_id = 'RHEL-08-020180' %}
+{%- set helperLoc = 'ash-linux/el8/STIGbyID/cat2/files' %}
+{%- set skipIt = salt.pillar.get('ash-linux:lookup:skip-stigs', []) %}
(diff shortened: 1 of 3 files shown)


ash-linux/el8/STIGbyID/cat2/RHEL-08-020180.sls
```diff
@@ -0,0 +1,46 @@
...
+    - stateful: True
+    - cwd: /root
+{%- else %}
+  {%- for user in userList %}
+Set minimum password lifetime for {{ user }}:
+  user.present:
+    - name: '{{ user }}'
+    - mindays: 1
+    - onlyif:
+      - '[[ -n $( awk -F: ''/{{ user }}:/ && $4 < 1  {print $1 " " $4}'' /etc/shadow ) ]]'
```
eemperor, line 44: Did you check that this does what it's suppose to?  It seems the mix of quote may not resolve correctly?

ferricoxide: Yes, it does exactly what it's supposed to do. Anticipating such questions is why I posted the test setup and relevant execution-outputs to our Slack channel.

Remember that, with the Pythonic processing of the Saltstack states:

- Standalone `'` is interpreted as a normal, string-bounding marker (for Line 44, that means the opening and ending `'`s)
- A double  `'` is essentially an escape that gets interpreted as a literal `'` within the bounded-string.

Which is to say that you end up with line 44 getting rendered as:

```
[[ -n $( awk -F: '/{{ user }}:/ && $4 < 1  {print $1 " " $4}' /etc/shadow ) ]]
```

…Which sets an "only execute this state's actions if the `/etc/shadow` file's field number 4 is null or `0` for the specified user" condition on the state.

eemperor (approved):

PR #379, opened 2023-04-20
Add handlers for V-230370

### LINKED ISSUE #369: [Feature Request] Passwords For New Users Must Have A Minimum Length Of 15 Characters

Per V-230370/RHEL-08-020231, passwords for new (RHEL8) users must have a minimum length of 15 characters.

Update the `PASS_MIN_LEN` value in the `/etc/login.defs` file to be `15`.

**Additional context**

ferricoxide:
Sets `PASS_MIN_LEN` value to `15` in the `/etc/login.defs` file

Closes #369

### DIFF
--- a/ash-linux/el8/STIGbyID/cat2/RHEL-08-020231.sls
+++ b/ash-linux/el8/STIGbyID/cat2/RHEL-08-020231.sls
@@ -1,0 +1,45 @@
+
+# Ref Doc:    STIG - RHEL 8 v1r9
+# Finding ID: V-230370
+# Rule ID:    SV-230370r627750_rule
+# STIG ID:    RHEL-08-020231
+# SRG ID:     SRG-OS-000078-GPOS-00046
+#
+# Finding Level: medium
+#
+# Rule Summary:
+#       Passwords for new users must have a minimum of 15 characters
+#
+# References:
+#   CCI:
+#     - CCI-000205
+#   NIST SP 800-53 :: IA-5 (1) (a)
+#   NIST SP 800-53A :: IA-5 (1).1 (i)
+#   NIST SP 800-53 Revision 4 :: IA-5 (1) (a)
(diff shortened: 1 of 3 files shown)


ash-linux/el8/STIGbyID/cat2/RHEL-08-020231.sls
```diff
@@ -0,0 +1,45 @@
...
+  cmd.run:
+    - name: 'printf "\nchanged=no comment=''Handler for {{ stig_id }} has been selected for skip.''\n"'
+    - stateful: True
+    - cwd: /root
+{%- else %}
+file_{{ stig_id }}-{{ targFile }}_poolDirectives:
+  file.replace:
+    - name: '{{ targFile }}'
+    - append_if_not_found: True
+    - pattern: '(^PASS_MIN_LEN)\s*[0-9]*.*$'
```
eemperor, line 43: Could there potentially be spaces at the beginning?

So by not using `#` in the search pattern, are you intending to leave the previous commented out setting alone and append the new setting?

ferricoxide: Technically, there _could_ be a space at the beginning …but only if something else (either a human or a bad set of automation) had previously-modified the file.

So, yeah, I'd debated making that `(^\s*PASS_MIN_LEN)\s*[0-9]*.*$'` or even `^(|\s*)(PASS_MIN_LEN)\s*[0-9]*.*$'`, but ultimately opted not to: if someone's running shit that fucks up that file _before_ (re)running watchmaker, they deserve what they get. :p

eemperor: Ok.  I'm good with it.

eemperor (approved):

PR #391, opened 2023-04-28
Add handlers for RHEL-08-010161

### LINKED ISSUE #359: [Feature Request] No KRB5 `.keytab` Files May Exist Unless `sssd` Created It

Per [RHEL-08-010161/V-230238](https://www.stigviewer.com/stig/red_hat_enterprise_linux_8/2020-11-25/finding/V-230238) "_RHEL 8 must prevent system daemons from using Kerberos for authentication_". However, for RHEL systems that are using the `sssd` service to authenticate to Kerberos-enabled sources (like Active Directory), an `/etc/krb5.keytab` file _must_ exist. 

Therefore, it will be necessary to add a conditional handler (under [`ash-linux/el8/STIGbyID/cat2`](https://github.com/plus3it/ash-linux-formula/tree/master/ash-linux/el8/STIGbyID/cat2)) to search for and delete any/all `*.keytab` files, but ensure to preserve the  `/etc/krb5.keytab` file if the `sssd` service is using it to authenticate to Kerberos-enabled sources.

Nuke all Kerberos `*.keytab` files unless the `sssd` service is using it to authenticate to Kerberos-enabled sources.

**Additional context**

ferricoxide:
Looks for and deletes any `.keytab` files in `/etc` if neither the `krb5-workstation` nor the `krb5-serve` RPMs is installed.

Closes #359

### DIFF
--- a/ash-linux/el8/STIGbyID/cat2/RHEL-08-010161.sls
+++ b/ash-linux/el8/STIGbyID/cat2/RHEL-08-010161.sls
@@ -1,0 +1,48 @@
+# Ref Doc:    STIG - RHEL 8 v1r9
+# Finding ID: V-230238
+# Rule ID:    SV-230238r646862_rule
+# STIG ID:    RHEL-08-010161
+# SRG ID:     SRG-OS-000120-GPOS-00061
+#
+# Finding Level: medium
+#
+# Rule Summary:
+#       The operating system must prevent system daemons from using
+#       Kerberos for authentication
+#
+# References:
+#   CCI:
+#     - CCI-000803
+#   NIST SP 800-53 :: IA-7
+#   NIST SP 800-53A :: IA-7.1
+#   NIST SP 800-53 Revision 4 :: IA-7
+#
+###########################################################################
+{%- set stig_id = 'RHEL-08-010161' %}
+{%- set helperLoc = 'ash-linux/el8/STIGbyID/cat2/files' %}
+{%- set skipIt = salt.pillar.get('ash-linux:lookup:skip-stigs', []) %}
+{%- set keytabList = salt.file.find('/etc', maxdepth=1, type='f', name='*.keytab') %}
+
+script_{{ stig_id }}-describe:
+  cmd.script:
+    - source: salt://{{ helperLoc }}/{{ stig_id }}.sh
+    - cwd: /root
+
+{%- if stig_id in skipIt %}
+notify_{{ stig_id }}-skipSet:
+  cmd.run:
+    - name: 'printf "\nchanged=no comment=''Handler for {{ stig_id }} has been selected for skip.''\n"'
+    - stateful: True
+    - cwd: /root
+{%- else %}
+  {%- for keytab in keytabList %}
+Delete suspect {{ keytab }} file:
+  file.absent:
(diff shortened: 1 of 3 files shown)


ferricoxide:
Unfortunately,  this::

~~~
Delete suspect {{ keytab }} file:
  file.absent:
    - name: '{{ keytab }}'
    - unless:
      - fun: pkg.version
        args:
          - krb5-workstation
          - krb5-server
~~~
Seems to be returning true:

~~~
[root@ip-0A004CD7 ~]# salt-call -c /opt/watchmaker/salt state.apply ash-linux.el8.STIGbyID.cat2.RHEL-08-010161
local:
----------
          ID: script_RHEL-08-010161-describe
    Function: cmd.script
      Result: True
     Comment: Command 'script_RHEL-08-010161-describe' run
     Started: 14:58:33.529356
    Duration: 69.185 ms
     Changes:
              ----------
              pid:
                  8151
              retcode:
                  0
              stderr:
              stdout:
                  --------------------------------------
                  STIG Finding ID: V-248543
                       The OS system must prevent system
                       daemons from using Kerberos for
                       authentication
                  --------------------------------------
----------
          ID: Delete suspect /etc/krb5.keytab file
    Function: file.absent
        Name: /etc/krb5.keytab
      Result: True
     Comment: unless condition is true
     Started: 14:58:33.639022
    Duration: 4867.891 ms
     Changes:

Summary for local
------------
Succeeded: 2 (changed=1)
Failed:    0
------------
Total states run:     2
Total run time:   4.937 s
~~~
regardless of presence/absence of the `krb5-workstaiton` and/or `krb5-server` RPMs:
~~~
[root@ip-0A004CD7 ~]# rpm -qa krb5-\*
krb5-libs-1.18.2-22.el8_7.x86_64
~~~
Which is to show that, while `krb5-libs` RPM is present on the test-system, the targeted RPMs are not.

This state _should_ be effectively superfluous: `/etc/*.keytab` files are only present on a to-be-hardened system when it's been domain-joined via `sssd` …and are automatically removed if one does a `realm leave` on an already-joined system. But, since I explicitly tested this state by doing `touch /etc/krb5.keytab` on an execution-target and it still skipped the removal-attempt, this control wants fine-tuning before being merged.

ash-linux/el8/STIGbyID/cat2/RHEL-08-010161.sls
```diff
@@ -0,0 +1,46 @@
...
+    - stateful: True
+    - cwd: /root
+{%- else %}
+  {%- for keytab in keytabList %}
+Delete suspect {{ keytab }} file:
+  file.absent:
+    - name: '{{ keytab }}'
+    - onlyif:
+      - '[[ $( rpm --quiet -q krb5-workstation )$? -ne 0 ]]'
+      - '[[ $( rpm --quiet -q krb5-server )$? -ne 0 ]]'
```
lorengordon, line 44: There's a somewhat simpler construct that has the same effect... approving either way

```suggestion
    - onlyif:
      - '! rpm --quiet -q krb5-workstation'
      - '! rpm --quiet -q krb5-server'
```

lorengordon (approved):

PR #392, opened 2023-05-02
Exits with oscap return code instead of tee return code

lorengordon:


ferricoxide (commented):
Adding capture of PIPESTATUS looks good. 

…Just to verify: we only care whether `oscap` worked and not whether `tee` worked (since `${PIPESTATUS[0]}` _only_ checks the former and not the latter)?

lorengordon:
> Adding capture of PIPESTATUS looks good.
> 
> …Just to verify: we only care whether `oscap` worked and not whether `tee` worked (since `${PIPESTATUS[0]}` _only_ checks the former and not the latter)?

Yeah, I think so? I can't think of a scenario where we'd want to fail because `tee` failed, but `oscap` succeeded...

ferricoxide (approved):
Prior question answered.

PR #393, opened 2023-05-03
Uses el7 and el8 compatible method to redirect oscap output and preserve return code

lorengordon:
Using PIPESTATUS on el7 resulted in the error:

    /bin/bash: line 0: exit: 2[0]: numeric argument required


ferricoxide (approved):
Be soooooo glad to get off EL7. So many more useful queriable-ENVs with the EL8 version of BASH.

Issue #120, opened 2016-01-14
Need Better Solution for EXTn xattr Attribs


ferricoxide:
The fix for formula-issue #118 is a kludge. Need to look at writing an actual file.\* module for SaltStack as it does not currently look like such is on the SaltStack roadmap (no target release set in notation for saltstack/salt#27094)

ferricoxide:
This bug is over a year old and the referenced issue in SaltStack is still open. Further, EL6 is coming up on its final scheduled release prior to End of Sales Life. Will keep issue open for tracking, but this will likely end up a WONTFIX.

lorengordon:
@KevinPlus3 is looking into this now.

KevinPlus3:
file.managed now uses file.chattr and file.lsattr to allow one to handle attributes of files/directories on Linux system.  The branch has been merged into https://github.com/saltstack/salt.git on the develop branch.

ferricoxide:
OBE: EL6 no longer supported by vendor

PR #402, opened 2023-05-22
Remediate V-250314 (RHEL-07-020023)

### LINKED ISSUE #399: [Feature Request] Address V-250314 (RHEL-07-020023)

Recent system scans have flagged V-250314 (RHEL-07-020023). Looks like new, valid STIG-content for which watchmaker does not yet have handler.

Create handler to address V-250314 (RHEL-07-020023)

**Additional context**

ferricoxide:
Merging will close #399

### DIFF
--- a/ash-linux/el7/STIGbyID/cat2/RHEL-07-020023.sls
+++ b/ash-linux/el7/STIGbyID/cat2/RHEL-07-020023.sls
@@ -33,15 +33,17 @@ notify_{{ stig_id }}-skipSet:
     - stateful: True
     - cwd: /root
 {%- else %}
-  {%- if selState.get("enabled") == True %}
     {%- for checkFile in sudoerFiles %}
 Fixing {{ checkFile }} (per {{ stig_id }}):
   file.replace:
     - name: '{{ checkFile }}'
     - append_if_not_found: False
     - backup: False
+    - onlyif:
+      - fun: grains.get
+        args:
+          - selinux:enabled
     - pattern: '^%([a-z0-9]*)(\s+)([A-Z]*)=([/(][A-Za-z]*[)])\s+(?!(TYPE|ROLE)=[a-z_]*)\s*([A-Za-z:]*)$'
     - repl: '%\1\2\3=\4 TYPE=sysadm_t ROLE=sysadm_r \6'
     {%- endfor %}
-  {%- endif %}
 {%- endif %}
--- a/ash-linux/el7/STIGbyID/cat2/RHEL-07-020023.sls
+++ b/ash-linux/el7/STIGbyID/cat2/RHEL-07-020023.sls
@@ -39,7 +39,7 @@ Fixing {{ checkFile }} (per {{ stig_id }}):
     - name: '{{ checkFile }}'
     - append_if_not_found: False
     - backup: False
-    - pattern: '^%([a-z0-9]*)(\s*\s*)([A-Z]*)=([/(][A-Za-z]*[)])\s\s*(?!(TYPE|ROLE)=[a-z_]*)([A-Za-z:]*)$'
+    - pattern: '^%([a-z0-9]*)(\s+)([A-Z]*)=([/(][A-Za-z]*[)])\s+(?!(TYPE|ROLE)=[a-z_]*)\s*([A-Za-z:]*)$'
     - repl: '%\1\2\3=\4 TYPE=sysadm_t ROLE=sysadm_r \6'
   {%- endfor %}
 {%- endif %}
--- a/ash-linux/el7/STIGbyID/cat2/RHEL-07-020023.sls
+++ b/ash-linux/el7/STIGbyID/cat2/RHEL-07-020023.sls
@@ -33,7 +33,8 @@ notify_{{ stig_id }}-skipSet:
     - stateful: True
     - cwd: /root
 {%- else %}
-  {%- for checkFile in sudoerFiles %}
+  {%- if selState.get("enabled") == True %}
+    {%- for checkFile in sudoerFiles %}
 Fixing {{ checkFile }} (per {{ stig_id }}):
   file.replace:
     - name: '{{ checkFile }}'
@@ -41,5 +42,6 @@ Fixing {{ checkFile }} (per {{ stig_id }}):
     - backup: False
     - pattern: '^%([a-z0-9]*)(\s+)([A-Z]*)=([/(][A-Za-z]*[)])\s+(?!(TYPE|ROLE)=[a-z_]*)\s*([A-Za-z:]*)$'
     - repl: '%\1\2\3=\4 TYPE=sysadm_t ROLE=sysadm_r \6'
-  {%- endfor %}
+    {%- endfor %}
+  {%- endif %}
 {%- endif %}
--- a/ash-linux/el7/STIGbyID/cat2/RHEL-07-020023.sls
+++ b/ash-linux/el7/STIGbyID/cat2/RHEL-07-020023.sls
@@ -26,9 +26,20 @@
   {%- do sudoerFiles.append(moreSudoerFiles) %}
 {%- endfor %}
 
-
-{%- for checkFile in sudoerFiles %}
-Checking {{ checkFile }} (per {{ stig_id }}):
+{%- if stig_id in skipIt %}
+notify_{{ stig_id }}-skipSet:
   cmd.run:
-    - name: 'grep % {{ checkFile }}'
-{%- endfor %}
+    - name: 'printf "\nchanged=no comment=''Handler for {{ stig_id }} has been selected for skip.''\n"'
+    - stateful: True
+    - cwd: /root
+{%- else %}
+  {%- for checkFile in sudoerFiles %}
+Fixing {{ checkFile }} (per {{ stig_id }}):
+  file.replace:
+    - name: '{{ checkFile }}'
+    - append_if_not_found: False
+    - backup: False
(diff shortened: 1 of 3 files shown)


ferricoxide:
Note: while the content in this PR _does_ result in `/etc/sudoers` and `/etc/sudoers.d/*` entries having the requested SELinux `TYPE` and `ROLE` tokens applied, doing so breaks the effected users:

```
[testuser@ip-0a004ede ~]$ sudo -i
sudo: unconfined_u:sysadm_r:sysadm_t:s0-s0:c0.c1023 is not a valid context
```

Need to do further investigation to resolve this issue.

ferricoxide:
Above issue appears to be related to the (long since de-supported) PBIS module still being used as the default AD-integration tool for EL7 EC2s in the lab. Swapping out PBIS (`domainjoin-cli leave`) for `sssd` seems to allow both continued AD-integration _and_ for SELinux mapped-users to `sudo`.

Probably going to need to add a state that sets the `__default__` SELinux user-context to `staff_u` or similar.

ferricoxide:
Note: this state-addition should probably not be added until [AMIgen7 Issue #120](https://github.com/plus3it/amigen7/issues/120) has been fully-exercised and merged in.

ash-linux/el7/STIGbyID/cat2/RHEL-07-020023.sls
```diff
@@ -0,0 +1,47 @@
...
+{%- endfor %}
+
+{%- if stig_id in skipIt %}
+notify_{{ stig_id }}-skipSet:
+  cmd.run:
+    - name: 'printf "\nchanged=no comment=''Handler for {{ stig_id }} has been selected for skip.''\n"'
+    - stateful: True
+    - cwd: /root
+{%- else %}
+  {%- if selState.get("enabled") == True %}
```
lorengordon, line 36: `selState` isn't defined anywhere, so this won't work as-is. But also, this would be skipped if selinux were disabled when Jinja executed, but then enabled by the scap profile. Probably better to use `unless` or `onlyif`.

ash-linux/el7/STIGbyID/cat2/RHEL-07-020023.sls
```diff
@@ -0,0 +1,47 @@
...
+{%- set sudoerFiles = [ '/etc/sudoers' ] %}
+{%- for moreSudoerFiles in salt.file.find('/etc/sudoers.d', type='f', grep='^%') %}
+  {%- do sudoerFiles.append(moreSudoerFiles) %}
+{%- endfor %}
+
+{%- if stig_id in skipIt %}
+notify_{{ stig_id }}-skipSet:
+  cmd.run:
+    - name: 'printf "\nchanged=no comment=''Handler for {{ stig_id }} has been selected for skip.''\n"'
+    - stateful: True
```
lorengordon, line 33: Marked stateful, but script is not setup for "no changes"

ash-linux/el7/STIGbyID/cat2/RHEL-07-020023.sls
ferricoxide: Dammit... `selState` _was_ defined: it was done via a `salt.grains.get('selinux')`. Where the hell did it go???

Must have gotten lost in the back-n-forth copy-n-paste from my authoring environment and the test-execution EC2. 😠

ferricoxide: > But also, this would be skipped if selinux were disabled when Jinja executed, but then enabled by the scap profile. 

Notionally not an issue: if SELinux is disabled, it requires a reboot for the state-change to take effect. Which is to say, within a single run of the Saltstack formula, the state is never going to change (or be able to change). While one can change from `Enforcing` to `Permissive` (or vice versa), one cannot change from Enabled to Disabled (or vice versa) without incurring a reboot.

Part of why I'd chosen the (apparently missing) `salt.grains.get('selinux')` method is that it includes a `.get()`able attribute:

```
# salt-call -c /opt/watchmaker/salt grains.get selinux
local:
    ----------
    enabled:
        True
    enforced:
        Enforcing
```

And I was pulling from the `enabled`  attribute …and chose _that_ in hopes that grain-content would be less "expensive". I could sub in either an:

```
onlyif:
  - '[[ $( getenforce | grep -qEi '(Enforcing|Permissive)' )$? -eq 0 ]]'
```
or a:
```
onlyif:
  - '[[ $( sestatus | grep SELinux\ status | sed 's/^.*:\s*//' ) == 'enabled' ]]'
```

I suppose? The latter being more analogous to the `selState.get()`.

lorengordon: `grains.get` is a salt function, so you could use the `fun:` feature of `onlyif` or `unless`, same as we did for other PRs.

ferricoxide: AH!

On it.

ferricoxide: Hrm... If I did:

```
    - onlyif:
      - fun: grains.get
        args:
          - selinux
```

That would just get me back the dict:

```
    enabled:
        True
    enforced:
        Enforcing
```

Not sure how to make that an actionable-equivalent of either the Jinja method or the shell-out methods

lorengordon: Yeah, everything you can do with `salt.<foo>` in jinja, you can do with the `fun:` method. Everything here is a salt function: https://docs.saltproject.io/en/latest/ref/modules/all/index.html

lorengordon: I think you can index into the dictionary using `:`, e.g. `'selinux:enabled'`

ferricoxide: Any way: mod _appears_ to work, but need to reboot to a SEL-disable state to validate it properly falses on that.

ferricoxide: Alright. When I reboot to SEL-disabled:

```
----------
          ID: Fixing /etc/sudoers.d/group_wheel (per RHEL-07-020023)
    Function: file.replace
        Name: /etc/sudoers.d/group_wheel
      Result: True
     Comment: onlyif condition is false
     Started: 18:07:33.061836
    Duration: 3.774 ms
     Changes:
```

Pushing the updates.

lorengordon:
Something went sideways when updating from master. The pr is showing a few of my commits that are already merged to master...

ferricoxide:
> Something went sideways when updating from master. The pr is showing a few of my commits that are already merged to master...

Yeah. I'd tried to rebase (because the GHAs were all stuck and I thought I was missing the bits that activated them). I reverted it.

lorengordon:
Were we going to hold off on merging this one until after we get confirmation that the known customers have switched from pbis to sssd?

ferricoxide:
> Were we going to hold off on merging this one until after we get confirmation that the known customers have switched from pbis to sssd?

Until _we_ update the per-environment salt-content archive-files, my assumption would be that _no one_ is going to be using `sssd`. Which is to say, I don't know how _any_ customers "switch" (except on a rebuild …where, presumably, they don't even know there's been a switch).

If the concern is people who might have a pet that they're attempting to re-execute watchmaker formula-content from – by replacing an older version of watchmaker with a newer one – I'm not sure if/how we ever solve that (edge) case.

lorengordon:
> Until _we_ update the per-environment salt-content archive-files, my assumption would be that _no one_ is going to be using `sssd`. Which is to say, I don't know how _any_ customers "switch" (except on a rebuild …where, presumably, they don't even know there's been a switch).
> 
> If the concern is people who might have a pet that they're attempting to re-execute watchmaker formula-content from – by replacing an older version of watchmaker with a newer one – I'm not sure if/how we ever solve that (edge) case.

Well, yes, and in the environments we manage, we have at least two that are stuck on pbis because we found issues when testing sssd. To avoid breaking those, I think we need to put a hold on this PR for now.

ferricoxide:
Note:

- [PR 121](https://github.com/plus3it/amigen7/pull/121) (AMIgen7) &ndash; already merged &ndash; changes the provisioning-user's SELinux confinement from `unconfined_u` to `staff_u`. Discovered that this created a need for the provisioning-user to modify their `sudo` execution to include a SELinux role- and type-request (e.g., `sudo -r unconfined_r -t unconfined_t` or `sudo -r sysadm_r -t sysadm_t`)
- [PR 123](https://github.com/plus3it/amigen7/pull/123) (AMIgen8) &ndash; pending approval as of this comment &ndash; addresses this behavior-change (ensuring that `sysadm_r` and `sysadm_t` are automatically-requested so that the user doesn't have to)

These doesn't solve the PBIS issues, just providing further awareness of relevant changes.

lorengordon:
> https://github.com/plus3it/amigen7/pull/123 (AMIgen8) – pending approval as of this comment – addresses this behavior-change (ensuring that sysadm_r and sysadm_t are automatically-requested so that the user doesn't have to)

Clarifying: That PR is on AMIgen7, not 8...

ferricoxide:
Both of the previously-noted AMIgen7 and AMIgen8 PRs have been merged. At this point, simply need _this_ PR &ndash; the enforcer-component for ASH-EL7 &ndash; to be approved for merge

lorengordon:
Confirmed the switch to sssd is complete in all environments. @ferricoxide anything left to investigate or test with this one?

lorengordon (approved):

PR #408, opened 2023-06-06
Split file- and dir-find actions

### LINKED ISSUE #407: [BUG] The `el7.STIGbyID.cat2.RHEL-07-020360` State Can Fail If Identified "Files" Are Actually Directories
**Describe the bug**

When initially authored in October of 2016, this state used a faulty method for identifying remediation-targets. The `find` operation that currently feeds the remediation-operation should have only been looking for files while excluding directories. Directories still need to be handled, but should be done in their own find/remediate state-couplet.

**Severity**

Edge-case: will cause a watchmaker-run to abort IFF something creates a `-nouser` directory before the (re)execution of this state.

**To Reproduce**
1. Launch a new system
2. Install software that creates directories with undefined-users as owners (the [forescout-secure-connector formula](https://github.com/plus3it/forescout-secure-connector-formula) is known to do this)
3. Execute the `el7.STIGbyID`/`el7.STIGbyID.cat2` content &ndash; or _just_ the `el7.STIGbyID.cat2.RHEL-07-020360` content
4. State will fail like:
    ```
    ----------
              ID: script_RHEL-07-020360-describe
        Function: cmd.script
          Result: True
         Comment: Command 'script_RHEL-07-020360-describe' run
         Started: 13:09:01.476940
        Duration: 156.11 ms
         Changes:   
                  ----------
                  pid:
                      8281
                  retcode:
                      0
                  stderr:
                  stdout:
                      ----------------------------------------
                      STIG Finding ID: RHEL-07-020360
                         All files and directories must have
                         a valid owner.
                      ----------------------------------------
    ----------
              ID: file_RHEL-07-020360-/usr/lib/forescout/bin/lib
        Function: file.managed
            Name: /usr/lib/forescout/bin/lib
          Result: False
         Comment: Specified target /usr/lib/forescout/bin/lib is a directory
         Started: 13:09:01.637496
        Duration: 5.181 ms
         Changes: 
    ```

**Expected behavior**

1. No errors should be emitted
2. `-nouser` files should be remediated 
3. `-nouser` directories should be remediated

**Additional context**

While the formula's remediable-content location-method is faulty, properly architected software-installations should not create files or directories with no valid owner defined.

**Fix Suggestions**

Update formula-contents to:
- Separate identification of `-nouser` files from `-nouser` directories
- Separate remediation-logic for `-nouser` files and `-nouser` directories
- Exchange current shell-out `find` method for SaltStack-native [`find`](https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.file.html) method
- Update STIG ID to match value from latest STIG &ndash; EL7 STIG v3R11 as of this issue's submission

ferricoxide:
The saltstack states for setting file and directory ownership are different, therefore, the object-finding list-creation logic needed to be tightened up. The STIG-handler now does one find for files with missing/unmapped user-ownership and one find for directories with missing/unmapped user-ownership.

Added "nothing found"  outputter to make things more-obvious that _something_ was executed to anyone reviewing logs

Could not convert to using `salt.file.find` since the native salt-module still doesn't (appear to) support the ability to look for `-nouser` attributes. 

Closes #407

### DIFF
--- a/ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls
+++ b/ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls
@@ -1,7 +1,9 @@
-# Finding ID:	RHEL-07-020360
-# Version:	RHEL-07-020360_rule
-# SRG ID:	SRG-OS-000480-GPOS-00227
-# Finding Level:	medium
+# Ref Doc:        STIG - RHEL 7 v3r11
+# Finding ID:     V-204463
+# STIG ID:	      RHEL-07-020320
+# Version:        RHEL-07-020320_rule
+# SRG ID:         SRG-OS-000480-GPOS-00227
+# Finding Level:  medium
 #
 # Rule Summary:
 #	All files and directories must have a valid owner.
@@ -10,7 +12,7 @@
 #    NIST SP 800-53 Revision 4 :: AC-3 (4)
 #
 #################################################################
-{%- set stig_id = 'RHEL-07-020360' %}
+{%- set stig_id = 'RHEL-07-020320' %}
 {%- set helperLoc = 'ash-linux/el7/STIGbyID/cat2/files' %}
 {%- set skipIt = salt.pillar.get('ash-linux:lookup:skip-stigs', []) %}
 {%- set nouserFiles = [] %}
--- a/ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls
+++ b/ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls
@@ -45,8 +45,22 @@ notify_{{ stig_id }}-skipSet:
     {%- set mountType = mountData[mount]['fstype'] %}
     {%- if mountData[mount]['fstype'] in localFstypes %}
       {%- set foundString = salt['cmd.shell']('find ' + mount + ' -xdev -type f -nouser') %}
-      {%- set foundList = foundString.split('\n') %}
-      {%- do nouserFiles.extend(foundList) %}
+      {%- if foundString %}
+        {%- set foundList = foundString.split('\n') %}
+        {%- do nouserFiles.extend(foundList) %}
+      {%- endif %}
+    {%- endif %}
+  {%- endfor %}
+
+# Find directories with no valid owner..
+  {%- for mount in mounts %}
+    {%- set mountType = mountData[mount]['fstype'] %}
+    {%- if mountData[mount]['fstype'] in localFstypes %}
+      {%- set foundString = salt['cmd.shell']('find ' + mount + ' -xdev -type d -nouser') %}
+      {%- if foundString %}
+        {%- set foundList = foundString.split('\n') %}
+        {%- do nouserDirs.extend(foundList) %}
+      {%- endif %}
     {%- endif %}
   {%- endfor %}
 
@@ -60,17 +74,13 @@ file_{{ stig_id }}-{{ file }}:
     - user: 'root'
       {%- endif %}
     {%- endfor %}
+  {%- else %}
+file_{{ stig_id }}-noneFound:
+  cmd.run:
+    - name: 'printf "\nchanged=no comment=''Found no files with missing/undefined users.''\n"'
+    - stateful: True
+    - cwd: /root
   {%- endif %}
-
-# Find directories with no valid owner..
-  {%- for mount in mounts %}
-    {%- set mountType = mountData[mount]['fstype'] %}
-    {%- if mountData[mount]['fstype'] in localFstypes %}
-      {%- set foundString = salt['cmd.shell']('find ' + mount + ' -xdev -type d -nouser') %}
-      {%- set foundList = foundString.split('\n') %}
-      {%- do nouserDirs.extend(foundList) %}
-    {%- endif %}
-  {%- endfor %}
 
 # Take ownership of directories
   {%- if nouserDirs|length %}
@@ -82,5 +92,11 @@ dir_{{ stig_id }}-{{ dir }}:
     - user: 'root'
       {%- endif %}
     {%- endfor %}
+  {%- else %}
+dir_{{ stig_id }}-noneFound:
+  cmd.run:
+    - name: 'printf "\nchanged=no comment=''Found no directories with missing/undefined users.''\n"'
+    - stateful: True
+    - cwd: /root
   {%- endif %}
 {%- endif %}
--- a/ash-linux/el7/STIGbyID/cat2/RHEL-07-020360.sls
+++ b/ash-linux/el7/STIGbyID/cat2/RHEL-07-020360.sls
@@ -14,6 +14,7 @@
 {%- set helperLoc = 'ash-linux/el7/STIGbyID/cat2/files' %}
 {%- set skipIt = salt.pillar.get('ash-linux:lookup:skip-stigs', []) %}
 {%- set nouserFiles = [] %}
+{%- set nouserDirs = [] %}
 {%- set localFstypes = [
                          'ext2',
                          'ext3',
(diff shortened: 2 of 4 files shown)


ferricoxide:
go codebuild go

ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls
```diff
@@ -0,0 +1,102 @@
...
+      {%- set foundString = salt['cmd.shell']('find ' + mount + ' -xdev -type d -nouser') %}
+      {%- if foundString %}
+        {%- set foundList = foundString.split('\n') %}
+        {%- do nouserDirs.extend(foundList) %}
+      {%- endif %}
+    {%- endif %}
+  {%- endfor %}
+
+# Take ownership of files
+  {%- if nouserFiles|length %}
```
lorengordon, line 68: The length filter isn't needed

```suggestion
  {%- if nouserFiles %}
```

ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls
```diff
@@ -0,0 +1,102 @@
...
+  {%- else %}
+file_{{ stig_id }}-noneFound:
+  cmd.run:
+    - name: 'printf "\nchanged=no comment=''Found no files with missing/undefined users.''\n"'
+    - stateful: True
+    - cwd: /root
+  {%- endif %}
+
+# Take ownership of directories
+  {%- if nouserDirs|length %}
```
lorengordon, line 86: The length filter isn't needed

```suggestion
  {%- if nouserDirs %}
```

ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls
```diff
@@ -0,0 +1,102 @@
...
+{%- if stig_id in skipIt %}
+notify_{{ stig_id }}-skipSet:
+  cmd.run:
+    - name: 'printf "\nchanged=no comment=''Handler for {{ stig_id }} has been selected for skip.''\n"'
+    - stateful: True
+    - cwd: /root
+{%- else %}
+# Find files with no valid owner..
+  {%- for mount in mounts %}
+    {%- set mountType = mountData[mount]['fstype'] %}
```
lorengordon, line 45: This isn't used. Can probably just remove it, since the value is only used once anyway.

```suggestion
```

ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls
```diff
@@ -0,0 +1,102 @@
...
+      {%- if foundString %}
+        {%- set foundList = foundString.split('\n') %}
+        {%- do nouserFiles.extend(foundList) %}
+      {%- endif %}
+    {%- endif %}
+  {%- endfor %}
+
+# Find directories with no valid owner..
+  {%- for mount in mounts %}
+    {%- set mountType = mountData[mount]['fstype'] %}
```
lorengordon, line 57: Unused, can remove

```suggestion
```

ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls
```diff
@@ -0,0 +1,102 @@
...
+    - user: 'root'
+      {%- endif %}
+    {%- endfor %}
+  {%- else %}
+file_{{ stig_id }}-noneFound:
+  cmd.run:
+    - name: 'printf "\nchanged=no comment=''Found no files with missing/undefined users.''\n"'
+    - stateful: True
+    - cwd: /root
+  {%- endif %}
```
lorengordon, line 83: I'm pretty sure this can be simplified using the `names` feature... Search for `names` in this doc: https://docs.saltproject.io/en/latest/ref/states/all/salt.states.file.html#salt.states.file.managed. See also: https://docs.saltproject.io/en/latest/ref/states/highstate.html#names-declaration

```suggestion
  {%- if nouserFiles %}
file_{{ stig_id }}-{{ file }}:
  file.managed:
    - names: {{ nouserFiles }}
    - user: 'root'
  {%- else %}
file_{{ stig_id }}-noneFound:
  cmd.run:
    - name: 'printf "\nchanged=no comment=''Found no files with missing/undefined users.''\n"'
    - stateful: True
    - cwd: /root
  {%- endif %}
```

ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls
```diff
@@ -0,0 +1,102 @@
...
+      {%- endif %}
+    {%- endfor %}
+  {%- else %}
+dir_{{ stig_id }}-noneFound:
+  cmd.run:
+    - name: 'printf "\nchanged=no comment=''Found no directories with missing/undefined users.''\n"'
+    - stateful: True
+    - cwd: /root
+  {%- endif %}
+{%- endif %}
```
lorengordon, line 102: Similarly, this can be simplified using `names`:

```suggestion
  {%- if nouserDirs %}
dir_{{ stig_id }}-{{ dir }}:
  file.directory:
    - names: {{ nouserDirs }}
    - user: 'root'
  {%- else %}
dir_{{ stig_id }}-noneFound:
  cmd.run:
    - name: 'printf "\nchanged=no comment=''Found no directories with missing/undefined users.''\n"'
    - stateful: True
    - cwd: /root
  {%- endif %}
{%- endif %}
```

ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls
```diff
@@ -0,0 +1,102 @@
...
+        {%- set foundList = foundString.split('\n') %}
+        {%- do nouserFiles.extend(foundList) %}
+      {%- endif %}
+    {%- endif %}
+  {%- endfor %}
+
+# Find directories with no valid owner..
+  {%- for mount in mounts %}
+    {%- set mountType = mountData[mount]['fstype'] %}
+    {%- if mountData[mount]['fstype'] in localFstypes %}
```
lorengordon, line 58: Can just loop over the map directly:

```suggestion
  {%- for mount, data in mountData.items() %}
    {%- if data['fstype'] in localFstypes %}
```

ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls
ferricoxide: Leftover from before I determined that my if/else was failing due to my list being a list of nulls rather than a null list

ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls
ferricoxide: Bah... Had originally had that line because was going to use it in the following `if` …until I realized I could more-directly evaluate.

ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls
ferricoxide: Ok, so, that may be _close_, but isn't quite it:

1. `file` is no longer defined
2.  It seemingly doesn't want to take a straight list:
    ```
    local:
        Data failed to compile:
    ----------
        Rendering SLS 'base:ash-linux.el7.STIGbyID.cat2.RHEL-07-020320' failed: while parsing a block mapping
      in "<unicode string>", line 27, column 7:
            - names: '['/root/blah_2', '/root/ ...
              ^
    expected <block end>, but found '<scalar>'
      in "<unicode string>", line 27, column 17:
            - names: '['/root/blah_2', '/root/blah_1']'
    ```

Digging around, now, for the proper construction.

ferricoxide: So, it _looks like_ it wants to be:
```
file_{{ stig_id }}:
  file.managed:
    - names:
    {%- for file in nouserFiles %}
      - '{{ file }}'
    {%- endfor %}
    - user: 'root'
 ```
Which actually aligns a _lot_ more with how Ansible does similar (though, Ansible has a `loop`  attribute that youd put `nouserFiles` into rather than having to resort to Jinja.

Comment (lorengordon) [reply ash-linux/el7/STIGbyID/cat2/RHEL-07-020320.sls]: this:

```
    - names:
    {%- for file in nouserFiles %}
      - '{{ file }}'
    {%- endfor %}
```

ought to be literally the same as:

```
    - names: {{ nouserFiles }}
```

the error with the string `        - names: '['/root/blah_2', '/root/blah_1']'` makes it look like you were stringifying the yaml list by wrapping it in outer quotes:

```
    - names: '{{ nouserFiles }}'
```

which indeed would be invalid syntax. worst case, just pass the value through the yamlencode filter:

```
    - names: {{ nouserFiles | yamlencode }}
```

but definitely do not stringify it!

ferricoxide:
Probably shoulda made my last commit-message "tersify"  rather than (another) "simplify".

lorengordon (approved):

PR #427, opened 2023-07-03
Add remediation for RHEL-08-040090

### LINKED ISSUE #422: [ENHANCEMENT] Add handler for EL8 V-230504
Per the STIG rule-summary:

> EL 8 firewall must employ a deny-all, allow-by-exception policy for allowing connections to other systems.

Currently, the firewall service uses the build-default zone of `public`. Need to add a handler to change this to `drop` (and ensure that the SSH service is exempted).

ferricoxide:
It _appears_ that the OSCAP content already takes care of this, but this will still show up on _some_ hardening-scans. Adding this state to make it easier to point auditors to a block of code that easily tells them, "yes, watchmaker _is_ setting this, appropriately"

Closes #422

### DIFF
--- a/ash-linux/el8/STIGbyID/cat2/RHEL-08-040090.sls
+++ b/ash-linux/el8/STIGbyID/cat2/RHEL-08-040090.sls
@@ -60,7 +60,7 @@ Set Default firewalld zone - config-running:
 Set Minimum Ports:
   firewalld.present:
     - name: '{{ firewalldValu }}'
-    - requires:
+    - require:
       - module: 'Set Default firewalld zone - config-running'
     - ports: {{ firewalldSafePorts }}
     - prune_ports: False
--- a/ash-linux/el8/STIGbyID/cat2/RHEL-08-040090.sls
+++ b/ash-linux/el8/STIGbyID/cat2/RHEL-08-040090.sls
@@ -60,7 +60,7 @@ Set Default firewalld zone - config-running:
 Set Minimum Ports:
   firewalld.present:
     - name: '{{ firewalldValu }}'
-    - onchanges:
+    - requires:
       - module: 'Set Default firewalld zone - config-running'
     - ports: {{ firewalldSafePorts }}
     - prune_ports: False
--- a/ash-linux/el8/STIGbyID/cat2/RHEL-08-040090.sls
+++ b/ash-linux/el8/STIGbyID/cat2/RHEL-08-040090.sls
@@ -1,0 +1,69 @@
+# Ref Doc:    STIG - RHEL 8 v1r10
+# Finding ID: V-230238
+# Rule ID:    SV-230504r854047_rule
+# STIG ID:    RHEL-08-040090
+# SRG ID:     SRG-OS-000297-GPOS-00115
+#
+# Finding Level: medium
+#
+# Rule Summary:
+#       The operating system must enable a firewall service that  employs
+#       a deny-all, allow-by-exception policy for allowing connections to
+#       other systems.
+#
+# References:
+#   CCI:
+#     - CCI-002314
+#   NIST SP 800-53 Revision 4 :: AC-17 (1)
+#
+###########################################################################
+{%- set stig_id = 'RHEL-08-040090' %}
+{%- set helperLoc = 'ash-linux/el8/STIGbyID/cat2/files' %}
+{%- set skipIt = salt.pillar.get('ash-linux:lookup:skip-stigs', []) %}
+{%- set firewalldConf = '/etc/firewalld/firewalld.conf' %}
+{%- set firewalldParm = 'DefaultZone' %}
+{%- set firewalldValu = salt.pillar.get('ash-linux:lookup:def_firewall_zone', 'drop') %}
+{%- set firewalldSafePorts = salt.pillar.get('ash-linux:lookup:def_firewall_ports', [] ) %}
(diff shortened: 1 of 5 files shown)


ferricoxide:
go codebuild go

ferricoxide:
go codebuild go

lorengordon (commented):
Curious about the choice of `onchanges`...? Seems the states ought to be safe to use just `requires`, due to `unless` for `module.run` and `firewalld.present` already being stateful?

ferricoxide:
Modifications to firewalld can happen to either the running (in memory) or on-disk changes. The two conditionals are meant to account for both:

* the `onchanges` detects that the on-disk has been changed
* the `unless` detects the in-memory (i.e., "active") configuration-status

Basically, tried to implement a "update the running configuration `IFF` there was a change on disk _and_ the current running-configuration is not the target state."

ash-linux/el8/STIGbyID/cat2/RHEL-08-040090.sls
```diff
@@ -0,0 +1,69 @@
...
+      - file: 'Set Default firewalld zone - config-file'
+    - unless:
+      - '[[ $( firewall-cmd --get-default-zone ) == "drop" ]]'
+    - zone: drop
+
+Set Minimum Ports:
+  firewalld.present:
+    - name: '{{ firewalldValu }}'
+    - onchanges:
+      - module: 'Set Default firewalld zone - config-running'
```
lorengordon, line 64: The risk here is that the running-config is already `drop`, and therefore the `module.run` state is skipped. But perhaps the "SafePorts" or "SafeSvcs" are absent from the firewalld config. Due to the `onchanges` requisite, this state would not run, so those configurations would not be made.

ash-linux/el8/STIGbyID/cat2/RHEL-08-040090.sls
```diff
@@ -0,0 +1,69 @@
...
+    - append_if_not_found: True
+    - not_found_content: |-
+        # Inserted per STIG {{ stig_id }}
+        {{ firewalldParm }}={{ firewalldValu }}
+
+Set Default firewalld zone - config-running:
+  module.run:
+    - name: firewalld.set_default_zone
+    - onchanges:
+      - file: 'Set Default firewalld zone - config-file'
```
lorengordon, line 55: The weird interaction here is if the config-file is already set to `drop`, but the config-running is not. Due to the `onchanges` requisite, this state would not execute, and so the config-running would not be corrected.

ferricoxide: If the config-file is already set to `drop`, the running-state sh/would also be `drop`. Typically, the only reason there would be a difference is if some other process was coming along and changing the running-state after it read the target state from disk. While it seems deranged in my mind that someone would do that, if they have a (presumably boot-time) process that's doing that, it would presumably be "dangerous" to change try to change the running-state (instead relying on the post-watchmaker reboot to push the system back to the hardening-mandated state).

ash-linux/el8/STIGbyID/cat2/RHEL-08-040090.sls
ferricoxide: I suppose that's a possibility. Change `onchanges` to `requires`?

ferricoxide: So, I guess this is a two-part/conditional query:

- Change the `onchanges` dependency to a `requires` dependency?

-or-

- Wholly dispense with trying to assert _any_ dependencies?

lorengordon: Either way, I think. Notionally using `requires` "feels" a little better, but if there isn't actually dependency on the order then it shouldn't matter.

ash-linux/el8/STIGbyID/cat2/RHEL-08-040090.sls
lorengordon: Ok, that's fine. I just wanted to better explain the question I had around the use of `onchanges` and the potential edge case.

lorengordon (dismissed):

ferricoxide:
Yeah. I saw the failure:

~~~

----------
          ID: Set Minimum Ports
    Function: firewalld.present
        Name: drop
      Result: False
     Comment: State 'firewalld.present' was not found in SLS 'ash-linux.el8.STIGbyID.cat2.RHEL-08-040090'
              Reason: 'firewalld' __virtual__ returned False: firewall-cmd is not available, firewalld is probably not installed.
~~~

But was tied up with other task to figure out why it didn't like that. I'll get to it next week.

ferricoxide:
Ah. Ok. Not so much:

> State 'firewalld.present' was not found in SLS 'ash-linux.el8.STIGbyID.cat2.RHEL-08-040090'

As:

> Reason: 'firewalld' __virtual__ returned False: firewall-cmd is not available, firewalld is probably not installed.

Guessing the VM-setup for the testing-container is missing the setup tasks that were in the EL7-based testing-containers.

ferricoxide:
Looks like `firewalld` was in the EL7-specific test-setup file but not EL8's. Since it needs to be in both &ndash; and will likely need to be in any future ELx platforms we need to support, I cut it from the EL7-specific test-setup and moved it to the common setup file.

PR #428, opened 2023-07-03
Add handler for V-230532 (RHEL-08-040180)

### LINKED ISSUE #421: [ENHANCEMENT] Add handler for EL8 V-230532
The relevant STIG discussion says:

> The debug-shell requires no authentication and provides root privileges to anyone who has physical access to the machine.  While this feature is disabled by default, masking it adds an additional layer of assurance that it will not be enabled via a dependency in systemd.  This also prevents attackers with physical access from trivially bypassing security on the machine through valid troubleshooting configurations and gaining root access when the system is rebooted.

While watchmaker is primarily geared towards virtualized systems (i.e., ones that have no "physical access"), similar access could be attained through virtualized-consoles like those available on VMware, Azure and other platforms. Therefore, need to add a handler to ensure the service is disabled (actually, per the STIG, "masked").

ferricoxide:
Add handler to set the `debug-shell.service`  systemd unit to `masked`.

Closes #421

### DIFF
--- a/ash-linux/el8/STIGbyID/cat2/RHEL-08-040180.sls
+++ b/ash-linux/el8/STIGbyID/cat2/RHEL-08-040180.sls
@@ -1,0 +1,41 @@
+# Ref Doc:    STIG - RHEL 8 v1r10
+# Finding ID: V-230532
+# Rule ID:    SV-230532r627750_rule
+# STIG ID:    RHEL-08-040180
+# SRG ID:     <none>
+#
+# Finding Level: medium
+#
+# Rule Summary:
+#       The debug-shell systemd service must be disabled.
+#
+# References:
+#   CCI:
+#     - CCI-000366
+#   NIST SP 800-53 :: CM-6 b
+#   NIST SP 800-53A :: CM-6.1 (iv)
+#   NIST SP 800-53 Revision 4 :: CM-6 b
+#
+###########################################################################
+{%- set stig_id = 'RHEL-08-040180' %}
+{%- set helperLoc = 'ash-linux/el8/STIGbyID/cat2/files' %}
+{%- set skipIt = salt.pillar.get('ash-linux:lookup:skip-stigs', []) %}
+
+script_{{ stig_id }}-describe:
+  cmd.script:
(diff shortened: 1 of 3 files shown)


ash-linux/el8/STIGbyID/cat2/RHEL-08-040180.sls
```diff
@@ -0,0 +1,41 @@
...
+{%- if stig_id in skipIt %}
+notify_{{ stig_id }}-skipSet:
+  cmd.run:
+    - name: 'printf "\nchanged=no comment=''Handler for {{ stig_id }} has been selected for skip.''\n"'
+    - stateful: True
+    - cwd: /root
+{%- else %}
+mask debug-shell service:
+  service.masked:
+    - name: "debug-shell.service"
```
lorengordon, line 39: Is there a scenario where this service doesn't exist?

ferricoxide: I'm thinking "probably not":
```bash
$ rpm --qf '%{name}\n' -qf $(
    systemctl status debug-shell.service | grep Loaded: | sed -e 's/^.*(//' -e 's/; .*$//'
)
systemd
```
and:
```bash
$ rpm -q --qf '%{name}\n' --whatrequires $(
    rpm --qf '%{name}\n' -qf $(
        systemctl status debug-shell.service | grep Loaded: | sed -e 's/^.*(//' -e 's/; .*$//'
    )
) | sort -u
acpid
audit
authselect-libs
chrony
cloud-init
cronie
dbus
device-mapper
dhcp-client
dracut
ec2-hibinit-agent
ec2-net-utils
firewalld
initscripts
iputils
kernel-core
libkcapi
microcode_ctl
NetworkManager
openssh-server
pcsc-lite
pcsc-lite-ccid
polkit
postfix
rng-tools
rsyslog
sssd-common
sssd-kcm
subscription-manager
systemd-pam
systemd-udev
timedatex
unbound-libs
usbguard
```
Basically, if that service is missing, then the `systemd` RPM has been very intentionally broken:

lorengordon (approved):

PR #429, opened 2023-07-03
Add handler for V-244533 (RHEL-08-020025)

### LINKED ISSUE #423: [ENHANCEMENT] Add handler for EL8 V-244533
Per the STIG V-ID's rule summery:

> RHEL 8 must configure the use of the pam_faillock.so module in the /etc/pam.d/system-auth file.

And discussion-text:

> By limiting the number of failed logon attempts, the risk of unauthorized system access via user password guessing, otherwise known as brute-force attacks, is reduced. Limits are imposed by locking the account.
>
> In RHEL 8.2 the "/etc/security/faillock.conf" file was incorporated to centralize the configuration of the pam_faillock.so module. Also introduced is a "local_users_only" option that will only track failed user authentication attempts for local users in /etc/passwd and ignore centralized (AD, IdM, LDAP, etc.) users to allow the centralized platform to solely manage user lockout.

Need to add a handler to ensure the presence of the `pam_faillock.so` module in the `/etc/pam.d/system-auth` file. Probably also want to dig into the documentation about the proper use of the `local_users_only` option (so that fat-fingers don't cause accounts to be locked both in the 3rd-party auth-source _and_ the EL8-based host

ferricoxide:
Ensures STIG-required `pam_faillock.so`  exist and exist in the proper stack-order(s)

Closes #423

### DIFF
--- a/ash-linux/el8/STIGbyID/cat2/RHEL-08-020025.sls
+++ b/ash-linux/el8/STIGbyID/cat2/RHEL-08-020025.sls
@@ -26,7 +26,6 @@
 {%- if salt.file.is_link(targFile) %}
   {%- set targFile = salt.cmd.run('readlink -f ' + targFile) %}
 {%- endif %}
-{%- set searchRoot = '^password\s+required\s+pam_pwhistory.so\s+' %}
 
 script_{{ stig_id }}-describe:
   cmd.script:
--- a/ash-linux/el8/STIGbyID/cat2/RHEL-08-020025.sls
+++ b/ash-linux/el8/STIGbyID/cat2/RHEL-08-020025.sls
@@ -55,7 +55,7 @@ Ensure auth pam_faillock.so preauth before pam_faillock.so authfail:
     - require:
       - file: 'Ensure auth pam_faillock.so authfail before pam_unix.so'
     - unless:
-      - '[[ $( grep -Pq "^auth\s*required\s*pam_faillock.so\s*preauth" {{ targFile }} )$? -eq 0 ]]'
+      - grep -Pq "^auth\s*required\s*pam_faillock.so\s*preauth" {{ targFile }}
 
 Ensure accunt pam_faillock.so before pam_unix.so:
   file.replace:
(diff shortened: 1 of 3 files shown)


ash-linux/el8/STIGbyID/cat2/RHEL-08-020025.sls
```diff
@@ -0,0 +1,69 @@
...
+    - stateful: True
+    - cwd: /root
+{%- else %}
+Ensure auth pam_faillock.so authfail before pam_unix.so:
+  file.replace:
+    - name: '{{ targFile }}'
+    - pattern: '(^auth\s*(sufficient|\[.*])\s*pam_unix.so.*)'
+    - repl: 'auth        required                                     pam_faillock.so authfail\n\1'
+    - unless:
+      - '[[ $( grep -Pq "^auth\s*required\s*pam_faillock.so\s*authfail" {{ targFile }} )$? -eq 0 ]]'
```
lorengordon, line 49: Can simplify:

```suggestion
      - grep -Pq "^auth\s*required\s*pam_faillock.so\s*authfail" {{ targFile }}
```

ash-linux/el8/STIGbyID/cat2/RHEL-08-020025.sls
```diff
@@ -0,0 +1,69 @@
...
+#     - CCI-000044
+#       - NIST SP 800-53 :: AC-7 a
+#       - NIST SP 800-53A :: AC-7.1 (ii)
+#       - NIST SP 800-53 Revision 4 :: AC-7 a
+#
+###########################################################################
+{%- set stig_id = 'RHEL-08-020025' %}
+{%- set helperLoc = 'ash-linux/el8/STIGbyID/cat2/files' %}
+{%- set skipIt = salt.pillar.get('ash-linux:lookup:skip-stigs', []) %}
+{%- set targFile = '/etc/pam.d/system-auth' %}
```
lorengordon, line 25: Do we have to consider a scenario where this file doesn't exist at all?

ferricoxide: No. If this file doesn't exist, the system is somewhere in the vicinity of "beyond broken". The files are installed as part of the PAM subsystem. No PAM, no RHEL:

```bash
$ rpm -q --qf '%{name}\n' --whatrequires $( rpm -qf /etc/pam.d/password-auth --qf '%{name}\n' )
util-linux
passwd
usermode
cronie
authselect-libs
openssh-server
```

lorengordon (approved):

PR #432, opened 2023-07-13
Add handler for V-244534 (RHEL-08-020026)

### LINKED ISSUE #424: [ENHANCEMENT] Add ash-linux-formula handler for EL8 V-244534
Per the STIG V-ID's rule summery:

> RHEL 8 must configure the use of the pam_faillock.so module in the /etc/pam.d/password-auth file.

And discussion-text:

> By limiting the number of failed logon attempts, the risk of unauthorized system access via user password guessing, otherwise known as brute-force attacks, is reduced. Limits are imposed by locking the account.
>
> In RHEL 8.2 the "/etc/security/faillock.conf" file was incorporated to centralize the configuration of the pam_faillock.so module.  Also introduced is a "local_users_only" option that will only track failed user authentication attempts for local users in /etc/passwd and ignore centralized (AD, IdM, LDAP, etc.) users to allow the centralized platform to solely manage user lockout.

Need to add a handler to ensure the presence of the `pam_faillock.so` module in the `/etc/pam.d/password-auth` file. Probably also want to dig into the documentation about the proper use of the `local_users_only` option (so that fat-fingers don't cause accounts to be locked both in the 3rd-party auth-source _and_ the EL8-based host

ferricoxide:
Ensures requisite `pam_faillock.so` entries in the `/etc/pam.d/password-auth` file

Closes #424

### DIFF
--- a/ash-linux/el8/STIGbyID/cat2/RHEL-08-020026.sls
+++ b/ash-linux/el8/STIGbyID/cat2/RHEL-08-020026.sls
@@ -63,5 +63,5 @@ Ensure accunt pam_faillock.so before pam_unix.so ({{ stig_id }}):
     - pattern: '(^account\s*required\s*pam_unix.so.*)'
     - repl: 'account     required                                     pam_faillock.so\n\1'
     - unless:
-      - '[[ $( grep -Pq "^account\s*required\s*pam_faillock.so" {{ targFile }} )$? -eq 0 ]]'
+      - grep -Pq "^account\s*required\s*pam_faillock.so" {{ targFile }}
 {%- endif %}
--- a/ash-linux/el8/STIGbyID/cat2/RHEL-08-020026.sls
+++ b/ash-linux/el8/STIGbyID/cat2/RHEL-08-020026.sls
@@ -45,7 +45,7 @@ Ensure auth pam_faillock.so authfail before pam_unix.so ({{ stig_id }}):
     - pattern: '(^auth\s*(sufficient|\[.*])\s*pam_unix.so.*)'
     - repl: 'auth        required                                     pam_faillock.so authfail\n\1'
     - unless:
-      - '[[ $( grep -Pq "^auth\s*required\s*pam_faillock.so\s*authfail" {{ targFile }} )$? -eq 0 ]]'
+      - grep -Pq "^auth\s*required\s*pam_faillock.so\s*authfail" {{ targFile }}
 
 Ensure auth pam_faillock.so preauth before pam_faillock.so authfail ({{ stig_id }}):
   file.replace:
--- a/ash-linux/el8/STIGbyID/cat2/RHEL-08-020026.sls
+++ b/ash-linux/el8/STIGbyID/cat2/RHEL-08-020026.sls
@@ -50,7 +50,7 @@ Ensure auth pam_faillock.so authfail before pam_unix.so ({{ stig_id }}):
 Ensure auth pam_faillock.so preauth before pam_faillock.so authfail ({{ stig_id }}):
   file.replace:
     - name: '{{ targFile }}'
-    - pattern: '(^auth\s*required\s*pam_faillock.so authfail)'
+    - pattern: '(^auth\s*required\s*pam_faillock.so\s*authfail)'
     - repl: 'auth        required                                     pam_faillock.so preauth\n\1'
     - require:
       - file: 'Ensure auth pam_faillock.so authfail before pam_unix.so ({{ stig_id }})'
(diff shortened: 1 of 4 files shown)


lorengordon (commented):
Is this one exactly the same as https://github.com/plus3it/ash-linux-formula/pull/429? What's the point of executing both?

ferricoxide:
This one addresses `/etc/pam.d/password-auth`, the other addresses `/etc/pam.d/system-auth`.  Technically separate findings so doing them as separate modules (rather than collapsing both into a single, list-iteration state/handler, mostly continuing to adhere to the previously-established convention of "one finding/one state-handler").

ash-linux/el8/STIGbyID/cat2/RHEL-08-020026.sls
```diff
@@ -0,0 +1,67 @@
...
+    - stateful: True
+    - cwd: /root
+{%- else %}
+Ensure auth pam_faillock.so authfail before pam_unix.so ({{ stig_id }}):
+  file.replace:
+    - name: '{{ targFile }}'
+    - pattern: '(^auth\s*(sufficient|\[.*])\s*pam_unix.so.*)'
+    - repl: 'auth        required                                     pam_faillock.so authfail\n\1'
+    - unless:
+      - '[[ $( grep -Pq "^auth\s*required\s*pam_faillock.so\s*authfail" {{ targFile }} )$? -eq 0 ]]'
```
lorengordon, line 48: I would just simplify these tests, same as the other pr...

```suggestion
      - grep -Pq "^auth\s*required\s*pam_faillock.so\s*authfail" {{ targFile }}
```

ash-linux/el8/STIGbyID/cat2/RHEL-08-020026.sls
```diff
@@ -0,0 +1,67 @@
...
+    - unless:
+      - '[[ $( grep -Pq "^auth\s*required\s*pam_faillock.so\s*preauth" {{ targFile }} )$? -eq 0 ]]'
+
+Ensure accunt pam_faillock.so before pam_unix.so ({{ stig_id }}):
+  file.replace:
+    - name: '{{ targFile }}'
+    - pattern: '(^account\s*required\s*pam_unix.so.*)'
+    - repl: 'account     required                                     pam_faillock.so\n\1'
+    - unless:
+      - '[[ $( grep -Pq "^account\s*required\s*pam_faillock.so" {{ targFile }} )$? -eq 0 ]]'
```
lorengordon, line 66:
```suggestion
      - grep -Pq "^account\s*required\s*pam_faillock.so" {{ targFile }}
```

ash-linux/el8/STIGbyID/cat2/RHEL-08-020026.sls
```diff
@@ -0,0 +1,67 @@
...
+    - name: '{{ targFile }}'
+    - pattern: '(^auth\s*(sufficient|\[.*])\s*pam_unix.so.*)'
+    - repl: 'auth        required                                     pam_faillock.so authfail\n\1'
+    - unless:
+      - grep -Pq "^auth\s*required\s*pam_faillock.so\s*authfail" {{ targFile }}
+
+Ensure auth pam_faillock.so preauth before pam_faillock.so authfail ({{ stig_id }}):
+  file.replace:
+    - name: '{{ targFile }}'
+    - pattern: '(^auth\s*required\s*pam_faillock.so authfail)'
```
lorengordon, line 53: Should this be?

```suggestion
    - pattern: '(^auth\s*required\s*pam_faillock.so\s*authfail)'
```

ferricoxide: _Possibly_ ...but, generally, once you reach the point where you're at the module and module-behavior modifiers, the whitespace is a single space rather than "zero or more white-spaces". Also, in this particular case, since the line being checked was previously inserted by line 46's `repl`, it's known that it will always be `… pam_faillock.so authfail`  and wouldn't be `… pam_faillock.so   authfail`. Basically, the immediately-prior state-execution _creates_ the landmark (since the two `pam_faillock.so` need to be before the `pam_unix` line but the `preauth` also needs to come before the `authfail` line). The method chosen for ensuring the correct order of the two, required lines seemed the most-reliable method.

lorengordon: I was mostly reviewing in the context of comparing to the patterns/greps in other states, which appeared to use `\s*`. But I may have been wrong; it's hard to review/understand what is happening in each of these steps, when just looking at the code and not playing around with running it on an actual system.

I suppose there could be a case where the file already has the entry via some other process, and that entry had multiple spaces. So, still seems like it's better to match the more general case.

lorengordon (approved):

PR #441, opened 2023-08-17
Allow fine-tuning of SELinux user-mappings

### LINKED ISSUE #440: [Feature Request] Add user-exclusions to `ash-linux/el7/STIGbyID/cat2/RHEL-07-020020.sls`

Some enterprise-integrations' installers do not properly create their service's `uid`s and/or `gid`s within the range reserved for "system" accounts (per the the ranges set within `/etc/login.defs` file's `SYS_UID_MIN`/`SYS_UID_MAX` and `SYS_GID_MIN`/`SYS_GID_MAX` parameter-pairs). As a result, when the `RHEL-07-020020` hardener runs, it treats those as normal, user accounts and asserts an SELinux user-confinement of `user_u`.

Extend the `RHEL-07-020020` hardener to be able to exempt a list of usernames from the remapping to the `user_u` SELinux confinement.

Get vendors to update their installers to do The Right Thing™ (creating their user-accounts within the ranges specified within `/etc/login.defs` file's `SYS_UID_MIN`/`SYS_UID_MAX` and `SYS_GID_MIN`/`SYS_GID_MAX` parameter-pairs.

**Additional context**

ferricoxide:
Adds the ability to read SELinux user-mapping associations from Pillar.

Closes #440

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -52,13 +52,13 @@
 
        # Users that should get specific confinements other that 'user_u'
 ##     sel_confine:
-##       guest-users:      # Users that should be mapped to the 'guest_u' confinement
-##       root-users:       # Users that should be mapped to the 'root' confinement
-##       staff-users:      # Users that should be mapped to the 'staff_u' confinement
-##       adm-users:        # Users that should be mapped to the 'sysadm_u' confinement
-##       system-users:     # Users that should be mapped to the 'system_u' confinement
-##       unconfined-users: # Users that should be mapped to the 'unconfined_u' confinement
-##       null-users:       # Users that should be not be mapped to a confinement
+##       guest_u:      # Users that should be mapped to the 'guest_u' confinement
+##       root_u:       # Users that should be mapped to the 'root' confinement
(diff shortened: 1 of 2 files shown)


pillar.example
```diff
@@ -48,6 +48,18 @@
        # Action to take if audispd is unable to send logs to a
        # remote collector-host
 ##     audisp-net-fail: <syslog|single|halt>
+#
+
+       # Users that should get specific confinements other that 'user_u'
+##     sel_confine:
```
eemperor, line 54: The example pillar has confinements that are not handled in the STIG handler above.  I'd say use examples that are actually processed or add the capability to process what's shown in the example.

Also, the pillar names in the example do not match the pillar names the STIG handler is looking for.  The `salt.pillar.get` commands are looking for `staff_u`, `sysadm_u`, and `unconfined_u` while the example has `staff-users`, `adm-users`, and `unconfined-users`.

ferricoxide: aligned.

eemperor (approved):

PR #462, opened 2023-10-12
Get rid of `exec` token

### LINKED ISSUE #459: [BUG] Guidance for EL8 `tmux` configuration has changed
**Describe the bug**

Earlier in the STIG lifecycle, the following configuration (in `/etc/profile.d/tmux` or equivalent) was prescribed:

~~~bash
if [ "$PS1" ]; then
  parent=$(ps -o ppid= -p $$)
  name=$(ps -o comm= -p $parent)
  case "$name" in (sshd|login) exec tmux ;; esac
fi
~~~

However, since the v1r9 release of the RHEL8 STIGs, the guidance was changed to:

~~~bash
if [ "$PS1" ]; then
  parent=$(ps -o ppid= -p $$)
  name=$(ps -o comm= -p $parent)
  case "$name" in (sshd|login) tmux ;; esac
fi
~~~

Leaving the former in place results in an (avoidable) scan error.

**Severity**

- [X] Moderately Broken (Trivial work-around)

**To Reproduce**
1. Deploy a RHEL 8 (or other EL8 work-alike) operating system
2. Apply project's current hardening-routines
3. Execute a STIG compliance-scan
4. Receive an out-of-compliance result from the `tmux`  configuration check

**Expected behavior**

Receive an in-compliance result from the `tmux`  configuration check

**Deviance Description**

Scans of `tmux`  configuration check should pass

**Screenshots**

**Additional context**

**Fix Suggestions**

Update `…/el8/STIGbyID/cat2/RHEL-08-020041.sls` to align to updated STIG-guidance

ferricoxide:
While we're here:

* Split out "is `tmux` available else quit" logic so scanners won't incorrectly mark the greater-logic as invalid
* Change to using (STIG-prescribed) `PS1` instead of checking if shell-attribute indicates interactive ...scanners won't incorrectly mark the logic as invalid
* Use the more-compressed (STIG-prescribed) `case` statement ...so scanners won't incorrectly mark them as invalid
* Remove other "human-friendly" formatting decisions ...so scanners won't incorrectly mark them as invalid

Closes #459

### DIFF
--- a/ash-linux/el8/STIGbyID/cat2/RHEL-08-020041.sls
+++ b/ash-linux/el8/STIGbyID/cat2/RHEL-08-020041.sls
@@ -46,7 +46,7 @@ file_{{ stig_id }}-{{ profileFile }}:
         # Check if tmux is available
         if [[ $( rpm -q tmux --quiet )$? -ne 0 ]] || [[ ! -x /usr/bin/tmux ]]
         then
-           exit
+           return
         fi
 
         # Check if shell is interactive
--- a/ash-linux/el8/STIGbyID/cat2/RHEL-08-020041.sls
+++ b/ash-linux/el8/STIGbyID/cat2/RHEL-08-020041.sls
@@ -1,4 +1,4 @@
-# Ref Doc:    STIG - RHEL 8 v1r7
+# Ref Doc:    STIG - RHEL 8 v1r9
 # Finding ID: V-230349
 # STIG ID:    RHEL-08-020041
 # Rule ID:    SV-230349r833388_rule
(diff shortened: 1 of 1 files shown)


ash-linux/el8/STIGbyID/cat2/RHEL-08-020041.sls
```diff
@@ -43,16 +43,15 @@ file_{{ stig_id }}-{{ profileFile }}:
...
     - dir_mode: '0755'
     - contents: |-
-        # Check if shell is interactive
-        if [[ $- == *i* ]] && [[ $( rpm --quiet -q tmux )$? -eq 0 ]]
+        # Check if tmux is available
+        if [[ $( rpm -q tmux --quiet )$? -ne 0 ]] || [[ ! -x /usr/bin/tmux ]]
         then
-           parent=$( ps -o ppid= -p $$ )
-           name=$( ps -o comm= -p $parent )
+           exit
```
lorengordon, line 49: Just checking that this logic got tested? I've run into weird situations around exiting a profile script in the past. Want to make sure this is behaving as intended. Is the goal to actually quit the _shell_ if tmux is not available? Or to skip the following `if` statement that runs tmux?

ferricoxide: Skip the following if. While it's not _as bad_ calling a binary that doesn't exist when you've removed the `exec` method, you still get an error-message if you attempt to invoke a binary that doesn't exist.

I'll go back and `chmod 0000` the `tmux` path to validate that it acts as expected.

ferricoxide: Yup. Forgot that the /etc/profile.d/* scripts are all `source`d (meaning that `exit` kills the sourcing-shell rather than terminating _just_ the `source`d script.

Updated.

lorengordon (approved):

PR #382, opened 2023-04-24
WIP: Reduce the States Showing As 'changed' During Highstate Execution

ferricoxide:
Changes calls to information-outputter scripts to be stateful and adds `changed=no` to referenced scripts.

Closes #380

PR #490, opened 2024-03-11
Logic to conditionally-create `user.cfg` files to support EFI-booted systems.

### LINKED ISSUE #487: [Feature Request] Ensure that UEFI Bootloader Password Is Configured On EFI-Enabled Systems
Existing content takes care of BIOS bootloader passwords but not UEFI bootloader password. Need to ensure each gets taken care of as appropriate.

image

While working this issue, maybe also ensure that admin-user declared in UEFI bootloader configuration is not the vendor-default.

ferricoxide:
Closes #487


ferricoxide:
image

ash-linux/el8/STIGbyID/cat1/RHEL-08-010140.sls
```diff
@@ -0,0 +1,82 @@
...
+    - stateful: True
+    - cwd: /root
+{%- else %}
+user_cfg_exists-{{ stig_id }}:
+  file.touch:
+    - name: '{{ grubPassFile }}'
+    - makedirs: True
+    - onlyif:
+      - [[ -d /sys/firmware/efi/ ]]
+      - [[ ! -e {{ grubPassFile }} ]]
```
lorengordon, line 54: This test isn't necessary, since the whole purpose of the `file.touch` state is to ensure the file exists, and by definition the state is stateful.

ash-linux/el8/STIGbyID/cat1/RHEL-08-010140.sls
```diff
@@ -0,0 +1,82 @@
...
+
+user_cfg_content-{{ stig_id }}:
+  cmd.run:
+    - name: 'printf "GRUB2_PASSWORD=%s\n" "$( printf "{{ grubPass }}\n{{ grubPass }}\n" | {{ grubUtil }} | awk ''/grub.pbkdf/{print $NF}'' )" > {{ grubPassFile }}'
```
lorengordon, line 58: Is there a way to make this stateful? Since the cmd is writing the whole file, I feel like we should be able to use `file.managed`. Run the command with jinja to get the contents, and use `file.managed` to write the file. That would also eliminate the need for the `file.touch` state.

ash-linux/el8/STIGbyID/cat1/RHEL-08-010140.sls
```diff
@@ -0,0 +1,82 @@
...
+
+grubuser_userSub-{{ grubUserFile }}-{{ stig_id }}:
+  file.replace:
+    - name: '{{ grubUserFile }}'
+    - pattern: 'password_pbkdf2 .* \\'
+    - repl: 'password_pbkdf2 {{ grubUser }} \\'
+
+regen_grubCfg-{{ stig_id }}:
+  cmd.run:
+    - name: '/sbin/grub2-mkconfig -o /boot/grub2/grub.cfg '
```
lorengordon, line 77: nit: random extra space at end of command?

```suggestion
    - name: '/sbin/grub2-mkconfig -o /boot/grub2/grub.cfg'
```

lorengordon, line 77: Also, since this command isn't stateful, it will run every time. Can we instead use `onchanges` so it only runs when the file changes?

ash-linux/el8/STIGbyID/cat1/RHEL-08-010140.sls
```diff
@@ -0,0 +1,82 @@
+# Ref Doc:    STIG - RHEL 8 v1r7
+# Finding ID: V-230235
+# STIG ID:    RHEL-08-010140
+# Rule ID:    SV-230235r743925_rule
+# SRG ID(s):  SRG-OS-000080-GPOS-00048
+# Finding Level:        high
+#
+# Rule Summary:
+#       RHEL 8 operating systems booted with a BIOS must
```
lorengordon, line 9: BIOS vs EFI?

```suggestion
#       RHEL 8 operating systems booted with EFI must
```

lorengordon (commented):
I think this isn't quite going to work the way we want... It will execute both `cat1.RHEL-08-010140` and `cat1.RHEL-08-010150` on every system, both EFI-booted and BIOS-booted systems. The `require` is used only for ordering, and only _cancels_ depending states if the required state _fails_. But "no changes" or "skip" are counted as "success". Which means the states operating on `grubUserFile`, which is the same file in each case, will execute twice.

ash-linux/el8/STIGbyID/cat1/RHEL-08-010140.sls
ferricoxide: Right now, the GRUB password is getting run twice for the same input-string. But, because each run of the tool uses a random seed, the string written to the BIOS `user.cfg` and the EFI `user.cfg` are different, even though they decrypt the same. Probably want to put the Jinja in a separate SLS file that each is able to `include`?

lorengordon: Hmm, if the output of the command is different every time it executes, then the file contents will always be changing. I guess maybe we could through a [`creates`](https://docs.saltproject.io/en/latest/ref/states/requisites.html#creates) requisite on it, so it doesn't run if the file already exists?

ferricoxide: Yup, different every time. Here's a quick example:

~~~bash
[root@ip-dqkjrhmqcez ~]# printf "test\ntest\n" | /bin/grub2-mkpasswd-pbkdf2 | awk '/grub.pbkdf2/{ print $NF }'
grub.pbkdf2.sha512.10000.DB4C41A2AC69B0706A9BBC89578D6DE6B8F309ADADE1F3BB05D13FD7B0831B3902412D3DF3DD7551B4C025767014BD0586286A0BCB306A85AD36EAC331836F9B.0F68FC8A3418DE4E0673F3F142BA6679AA2E1A962DB4FB48A0ACBC82E81162BB6750FF12337616B48335F706155FF31511A3D8D613AEF08347867899A65A3CB0

[root@ip-dqkjrhmqcez ~]# printf "test\ntest\n" | /bin/grub2-mkpasswd-pbkdf2 | awk '/grub.pbkdf2/{ print $NF }'
grub.pbkdf2.sha512.10000.89802A62E9C9E7CEDAC11D310D5BB9CFCCBA57487FDECB23F5D1D48510039C3C300A51CD981FBE4CB7447D4D12DC25E3E31D0BDBEBCEA4C75E0A85C0083A7228.377046676C1B61F8D6ABF5099B6868B829BEA3D5703A5FC4CC5B5B1D5B48DC7E5E444D7C312A5D5FC8B6027C414A1CB102613F594DC76DD050B8D4C4B6D4483E

[root@ip-dqkjrhmqcez ~]# printf "test\ntest\n" | /bin/grub2-mkpasswd-pbkdf2 | awk '/grub.pbkdf2/{ print $NF }'
grub.pbkdf2.sha512.10000.72101FC8DB6E5D72959800B9008EFEFD79435454D0B764C241A4F625E67C0DBA702180B69D1C530150266A720C4CF1A623726214B0482B7F612F22B2BCC40E36.302C411BF05C124ADCADB0EEFC45EFB76F4FE76C1381E6FE636CEE68C40A644C87EF344AE51454CB21CE212290303F70544A97967832B45B407FA01ED7B5C3B5

[root@ip-dqkjrhmqcez ~]# printf "test\ntest\n" | /bin/grub2-mkpasswd-pbkdf2 | awk '/grub.pbkdf2/{ print $NF }'
grub.pbkdf2.sha512.10000.A4E24387FFC78AB3565AE2AB240EEBB287EFC62C7718453094A1F9C6AF90782E2C174EA8AC81D8FFCBD8786D5C829809C1F86965320A8042739800A25B218911.5C0FC812236D7FAD3CEC210A739C377F161E2D72789F07514769338DE6487FFC38B99856CE55CBD644BCBBA4B518DB7923412FAB46B627348834114549C9E79B
~~~

ash-linux/el8/STIGbyID/cat1/RHEL-08-010140.sls
```diff
@@ -0,0 +1,85 @@
...
+{%- else %}
+  {%- set grubPassFile = '/boot/efi/EFI/redhat/user.cfg' %}
+{%- endif %}
+
+{%- set grubUtil = '/bin/grub2-mkpasswd-pbkdf2' %}
+
+script_{{ stig_id }}-describe:
+  cmd.script:
+    - source: salt://{{ helperLoc }}/{{ stig_id }}.sh
+    - cwd: /root
```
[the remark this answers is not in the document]
lorengordon: Oh you removed that comment in a subsequent commit, after i started looking at it...

ash-linux/el8/STIGbyID/cat1/RHEL-08-010140.sls
ferricoxide: Update the conditional logic: I only want `touch` to run if the file _doesn't_ exist. Using `touch`  rather than `managed` to ensure the file's existence `touch`  doesn't try to set modes or SEL labels.

ash-linux/el8/STIGbyID/cat1/RHEL-08-010140.sls
ferricoxide: Updated to pull the one-way hash from a .jinja file – and then referencing that file from both 140 and 150 – but still results in strings being different in the BIOS and EFI `user.cfg`  files. 

"Close".

☹️

ash-linux/el8/STIGbyID/cat1/RHEL-08-010140.sls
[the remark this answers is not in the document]
ferricoxide: Yup. And, per OoB conversation, switched from an outputter-script to `test.show_notification`. 😃

ash-linux/el8/STIGbyID/cat1/RHEL-08-010140.sls
```diff
@@ -0,0 +1,104 @@
...
+             user in order to access single-
+             user and maintenance modes
+        --------------------------------------
+
+{%- if stig_id in skipIt %}
+notify_{{ stig_id }}-skipSet:
+  cmd.run:
+    - name: 'printf "\nchanged=no comment=''Handler for {{ stig_id }} has been selected for skip.''\n"'
+    - stateful: True
+    - cwd: /root
```
lorengordon, line 50: Quibble, as this could also be `test.show_notification`

lorengordon (approved):

ash-linux/el8/STIGbyID/cat1/RHEL-08-010140.sls
ferricoxide: oop. hit merge before I scrolled back and noticed this.

lorengordon: Like I said, just a quibble, wasn't going to block merge on it. Maybe a note to add to the other PR, https://github.com/plus3it/ash-linux-formula/pull/382

PR #499, opened 2024-04-18
Add a 'SCAP undo' state

### LINKED ISSUE #498: [BUG] Upstream OSCAP Content Is (Currently) adding `*.* @@logcollector`  to `/etc/rsyslog.conf`
**Describe the bug**

When running the `oscap --remediate`  portion of the automation, the upstream OSCAP content is causing a `*.* @@logcollector`  entry to be `/etc/rsyslog.conf`. This results in spurious attempts to dupe system logs to a remote log-collector host that doesn't exist and created failure log-entries for those failed attempts

**Severity**

Nuisance

**To Reproduce**
1. Launch a spel or other ELn-based VM-template
1. Run the full watchmaker formula-set
1. After reboot, find error-messages about the `logcollector`  host not being found

**Expected behavior**

The `rsyslog`  service should _not_ be configured by the `ash-linux-formula`  content

**Deviance Description**

The `rsyslog`  service gets configured for remote-logging even though it should usually not be configured and _never_ via this formula.

**Screenshots**

**Additional context**

**Fix Suggestions**

Add some kind of post `oscap --remediate`  step to ensure that spuriously-created `*.* @@logcollector`  entries are removed from the `/etc/rsyslog.conf`  file.

ferricoxide:
Hopefully, the rsyslog "fix" logic is the only one we'll need (and, further hopefully, the content-maintainers accept and agree to fix the bug-repor about rsyslog and logcollector) to put here, but we now have a central place for adding such undos as necessary.

Closes #498

Note: if someone intentionally adds a `*.* @@logcollector` line to their host's `/etc/rsyslog.conf` file _before_ this state is run, this update will cause that entry to be commented out. That said, Red Hat generally recommends that remote-logging be set up in suitable `/etc/rsyslog.d/*` files (which this update wouldn't touch).

### DIFF
--- a/ash-linux/el8/VendorSTIG/scap_undos.sls
+++ b/ash-linux/el8/VendorSTIG/scap_undos.sls
@@ -11,4 +11,4 @@ undo logcollector in /etc/rsyslog.conf:
     - name: '/etc/rsyslog.conf'
     - not_found_content: ''
     - pattern: '^(\s*|#*\s*|)\*\.\*\s*@*logcollector$'
-    - repl: '# *.* @logcollector'
+    - repl: ''
--- a/ash-linux/el8/VendorSTIG/scap_undos.sls
+++ b/ash-linux/el8/VendorSTIG/scap_undos.sls
@@ -7,6 +7,8 @@
 # Undo SCAP's appending of a `*.* @@logcollector` config-token in the
 # rsyslog.conf file
 undo logcollector in /etc/rsyslog.conf:
-  file.comment:
+  file.replace:
     - name: '/etc/rsyslog.conf'
-    - regex: '^(\s*|)\*\.\*\s*@@logcollector'
+    - not_found_content: ''
+    - pattern: '^(\s*|#*\s*|)\*\.\*\s*@*logcollector$'
+    - repl: '# *.* @logcollector'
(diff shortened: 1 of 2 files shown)


ash-linux/el8/VendorSTIG/scap_undos.sls
```diff
@@ -0,0 +1,12 @@
+# This Salt state undoes anything put in place by the
+# "remediate" state that shouldn't have been done as part of the
+# generic OSCAP content
+#
+#################################################################
+
+# Undo SCAP's appending of a `*.* @@logcollector` config-token in the
+# rsyslog.conf file
+undo logcollector in /etc/rsyslog.conf:
+  file.comment:
```
lorengordon, line 10: Do we maybe want to use `file.replace` to just delete the matching line? That way the file doesn't accumulate comments on re-execution...

ferricoxide: Hadn't really considered top-to-bottom re-execution (given how disruptive it can be to a "configured for production" system). I'll go fix.

ash-linux/el8/VendorSTIG/scap_undos.sls
```diff
@@ -0,0 +1,14 @@
...
+#################################################################
+
+# Undo SCAP's appending of a `*.* @@logcollector` config-token in the
+# rsyslog.conf file
+undo logcollector in /etc/rsyslog.conf:
+  file.replace:
+    - name: '/etc/rsyslog.conf'
+    - not_found_content: ''
+    - pattern: '^(\s*|#*\s*|)\*\.\*\s*@*logcollector$'
+    - repl: '# *.* @logcollector'
```
lorengordon, line 14: Won't that still inject the comment?

```suggestion
    - repl: ''
```

lorengordon (approved):

Issue #496, opened 2024-04-10
[Feature Request] Add Support for RHEL9 and Related Distros


ferricoxide:
**Is your feature request related to a problem? Please describe.**

Efforts to be proactive in supporting migrations to RHEL 9, need to add hardening-support for RHEL 9 and related distros

**Describe the solution you'd like**

Add hardening-support for RHEL 9 and related distros

**Describe alternatives you've considered**

**Additional context**

ferricoxide:
As with #508, this issue is OBE due in large part because of addressing issues:
* #527
* #529
* #530
* #531
* #532
* #533
* #551
* #552
* #553
* #554
* #555
* #556
* #557
* #558
* #560
* #561
* #562
* #563
* #564
* #566
* #567
* #568
* #571
* #572
* #573
* #574
* #575
* #576
* #577
* #578
* #579
* #603
* #604

Issue #44, opened 2015-01-16
Handler for STIG-ID V-38497 is not reliably executing file existence-check in proper order


ferricoxide:
Placed logic inside SLS file to check for existence of system-auth-ac file:

    {% if not salt['file.file_exists'](checkFile) %}
    cmd_V38497-linkSysauth:
      cmd.run:
      - name: '/usr/sbin/authconfig --update'
    {% endif %}

If interpreted in order listed in SLS file, all is good. However, salt-call seems like it's doing out-of-order execution, rendering the above essentially useless. Need to force Salt to do the Right Thing.

PR #506, opened 2024-05-29
Adds EL9 hardening-content per Issue 496

ferricoxide:
Adds EL9 hardening-content per Issue #496


ferricoxide:
Ok, I think we're to the point – at least with CaC content – where we're hardened-up such that the rest of the stuff needs to be a watchmaker FAQ:
image
All of these are either "can't do this without breaking functionality", "can't do this for a global tool: you need to do this as a site-local modification" or "the scan-tests are not valid"

ferricoxide:
I _think_ the failing tests are because of CentOS Stream 8's discontinuance?

At any rate, this takes care of the technically-implementable remediations of `oscap`-identified findings (beyond what's mentioned in my prior comment). I'll do the `scc` content in a futher PR.

ferricoxide:
w00t! Review, plz, @plus3it/watchmaker!

ferricoxide:
@plus3it/watchmaker:

Bump.

ash-linux/el9/RuleById/medium/content_rule_firewalld_sshd_port_enabled.sls
```diff
@@ -0,0 +1,113 @@
...
+    - unless:
+      - grep -q {{ nic }}$ {{ nmcliFiles | join('/system-connections/* ')}}
+Set Zone for {{ nic }}:
+  module.run:
+    - name: firewalld.add_interface
+    - interface: {{ nic }}
+    - require:
+      - module: Enable SSHD globally
+    - require_in:
+      - test: Brake
```
lorengordon, line 73: Probably was here for testing? Not sure what the `Brake` state was about...

ash-linux/el9/RuleById/medium/content_rule_firewalld_sshd_port_enabled.sls
```diff
@@ -0,0 +1,113 @@
...
+  module.run:
+    - name: firewalld.add_service
+    - onlyif:
+      - fun: pkg.version
+        args:
+          - firewalld
+    - permanent: True
+    - require_in:
+      - module: Enable SSHD globally
+      - test: Brake
```
lorengordon, line 99: Delete if removing the `Brake` state...

ash-linux/el9/RuleById/medium/content_rule_mount_option_tmp_noexec.sls
```diff
@@ -0,0 +1,201 @@
...
+    - name: '{{ optionsDir }}'
+    - group: 'root'
+    - mode: '0700'
+    - selinux:
+        serange: 's0'
+        serole: 'object_r'
+        setype: 'systemd_unit_file_t'
+        seuser: 'system_u'
+    - unless:
+      - [[ -d {{ optionsDir }} ]]
```
lorengordon, line 160: This is extraneous, the state does this on its own.

ash-linux/el9/RuleById/medium/content_rule_mount_option_tmp_noexec.sls
```diff
@@ -0,0 +1,201 @@
...
+  file.replace:
+    - name: '{{ optionsFile }}'
+    - pattern: '^(Options=..*)'
+    - repl: '\1,{{ mntOpt }}'
+    - require:
+      - file: Create Dummy {{ optionsFile }} ({{ stig_id }})
+    - unless:
+      - 'grep -q ^Options=.*{{ mntOpt }} {{ optionsFile }}'
+      - file: Add first mount-option {{ mntOpt }} ({{ stig_id }})
+  {%- endfor %}
```
lorengordon, line 200: I'm struggling to understand what the end result of this is supposed to look like. It seems like this file is using INI syntax, so maybe we can simplify using an [`ini_manage` state](https://docs.saltproject.io/en/3004/ref/states/all/salt.states.ini_manage.html)?

ash-linux/el9/RuleById/medium/content_rule_postfix_client_configure_mail_alias.sls
```diff
@@ -0,0 +1,73 @@
...
+  file.replace:
+      - name: '{{ mailAliasFile }}'
+      - append_if_not_found: True
+      - not_found_content: |-
+
+          # Inserted per {{ stig_id }}
+          root\: {{ rootMailDest }}
+      - onlyif:
+        - '[[ -e {{ mailAliasFile }} ]]'
+      - pattern: '^([rR][oO][oO][tT]|\"[rR][oO][oO][tT]\")(\s*:\s*)(.+)$'
```
lorengordon, line 61: Salt does support python's regex flags, so could just make this a case-insensitive pattern, using `(?i...)`, see https://docs.python.org/3/library/re.html#re.I

ash-linux/el9/VendorSTIG/packages.sls
```diff
@@ -0,0 +1,17 @@
...
+{%- set helperLoc = tpldir ~ '/files' %}
+
+packages_{{ stig_id }}-installed:
+  pkg.installed:
+    - pkgs:
+      - openscap
+      - openscap-scanner
+{%- if salt.grains.get('os') == 'CentOS' %}
+      - openscap-engine-sce
+{%- endif %}
```
lorengordon, line 16: I know this was copied, but maybe don't need this CentOS check anymore?

ash-linux/el9/VendorSTIG/remediate.sls
```diff
@@ -0,0 +1,69 @@
...
+
+# Ensure root account password is configured to not expire
+root_password_no_expire:
+  user.present:
+    - name: root
+    - createhome: False
+    - mindays: -1
+    - maxdays: -1
+    - require:
+      - cmd: run_{{ stig_id }}-remediate
```
lorengordon, line 69: Now that we have the `scap_undos` sls, perhaps would make sense to move these there?

.github/dependabot.yml
```diff
@@ -1,6 +1,15 @@
...
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+  # Maintain dependencies for root dockerfile
+  - package-ecosystem: docker
+    directory: /
+    schedule:
+      interval: weekly
   - package-ecosystem: docker
```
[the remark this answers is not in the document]
ferricoxide: Removed per 580cf55

ash-linux/el9/RuleById/medium/content_rule_firewalld_sshd_port_enabled.sls
```diff
@@ -0,0 +1,113 @@
...
+    - zone: {{ targZone }}
+    {%- endif %}
+  {%- endfor %}
+
+##########################
+# Zone-configuration tasks
+##########################
+Brake:
+  test.show_notification:
+    - text: NO-OP
```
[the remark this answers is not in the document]
ferricoxide: Removed per 3e00d16

ash-linux/el9/RuleById/medium/content_rule_firewalld_sshd_port_enabled.sls
ferricoxide: Removed per 3e00d16

ash-linux/el9/RuleById/medium/content_rule_firewalld_sshd_port_enabled.sls
ferricoxide: Removed per 3e00d16

ash-linux/el9/RuleById/medium/content_rule_mount_option_tmp_noexec.sls
ferricoxide: Fixed in 868d0ade

ash-linux/el9/RuleById/medium/content_rule_mount_option_tmp_noexec.sls
ferricoxide: I chose the method I did mostly because the override file, at two lines, is short (content shown is as created as of today's RPM-delivered defaults for the `/usr/lib/systemd/system/tmp.mount`) . As it stands, its brevity is resultant of only needing to override a single option-value in a single configuration-block:

~~~
[Mount]
Options=mode=1777,strictatime,nosuid,nodev,size=50%%,nr_inodes=1m,noexec
~~~

Had the `/etc` file required a more fully fleshed-out INI-style configuration – requiring handling settings in multiple stanzas – I probably would have used `ini_manage`. Given its brevity, I hadn't really considered using anything but `file` states. 

The above payload is sufficient to override the `systemd` config-defaults (via the `systemd` RPM):

~~~
#  SPDX-License-Identifier: LGPL-2.1-or-later
#
#  This file is part of systemd.
#
#  systemd is free software; you can redistribute it and/or modify it
#  under the terms of the GNU Lesser General Public License as published by
#  the Free Software Foundation; either version 2.1 of the License, or
#  (at your option) any later version.

[Unit]
Description=Temporary Directory /tmp
Documentation=https://systemd.io/TEMPORARY_DIRECTORIES
Documentation=man:file-hierarchy(7)
Documentation=https://www.freedesktop.org/wiki/Software/systemd/APIFileSystems
ConditionPathIsSymbolicLink=!/tmp
DefaultDependencies=no
Conflicts=umount.target
Before=local-fs.target umount.target
After=swap.target

[Mount]
What=tmpfs
Where=/tmp
Type=tmpfs
Options=mode=1777,strictatime,nosuid,nodev,size=50%%,nr_inodes=1m

# Make 'systemctl enable tmp.mount' work:
[Install]
WantedBy=local-fs.target
~~~

ash-linux/el9/RuleById/medium/content_rule_postfix_client_configure_mail_alias.sls
ferricoxide: Dammit... Lemme go test.

ash-linux/el9/RuleById/medium/content_rule_mount_option_tmp_noexec.sls
lorengordon: I was commenting on the whole setup, the three state stanza... From what I'm seeing and what you're describing, here's what is going through my head as I try to parse the logic....

* If the file is delivered by rpm, `/usr/lib/systemd/system/tmp.mount`, then the `file.managed` state will never do anything due to the `unless` test. It won't even manage the mode, group, or user.
* If the file is not present, then `file.managed` executes, but doesn't do _enough_ to configure the unit file correctly?
* The logic for `set mountOptsDefault` assumes the file exists anyway, and will fail if it doesn't

Maybe instead of the "Dummy" file.managed state, just test if the file exists and if it does then proceed, otherwise, skip the whole sls?

Then the second two, I think all they're really doing is ensuring the `noexec` option is present? Or rather, some arbitrary number of "stig" options. I think that can be done by looping over _just_ `mountOptsStig` with the unless grep?

  {%- for mntOpt in mountOptsStig %}
Add extra mount-option {{ mntOpt }} ({{ stig_id }}):
  file.replace:
    - name: '{{ optionsFile }}'
    - pattern: '^(Options=.*)'
    - repl: '\1,{{ mntOpt }}'
    - require:
      - file: Create Dummy {{ optionsFile }} ({{ stig_id }})
    - unless:
      - 'grep -q ^Options=.*{{ mntOpt }} {{ optionsFile }}'
  {%- endfor %}

ferricoxide: > * If the file is delivered by rpm, /usr/lib/systemd/system/tmp.mount, then the file.managed state will never do anything due to the unless test. It won't even manage the mode, group, or user.
> * If the file is not present, then file.managed executes, but doesn't do enough to configure the unit file correctly?
> * The logic for set mountOptsDefault assumes the file exists anyway, and will fail if it doesn't

There will _never_ be a time – except on a horribly broken system – that the `/usr/lib/systemd/system/*` files won't exist. RPM dependencies won't allow it on a systemd-enabled ELn system. 

The whole point of the logic is to create the (two-line) _override_ file that lives in `/etc` while ensuring that all of the mount-opts found in the `/usr` file get incorporated into the `/etc/` file _plus_ all of the STIG-mandated mount-opts. The override file is short because anything not specified in the `/etc/` file is left as set in the `/usr` file and individual setting specified in the `/etc` file wholly overrides the setting from the `/usr` file (thus needing to capture the delivered options while adding the STIG-mandated options)

ash-linux/el9/VendorSTIG/packages.sls
ferricoxide: I need to go see whether the above made the run against a CentOS Stream 9 EC2 vomit…

ferricoxide: N/m, right now can't even run this formula on CentOS Stream 9:

~~~
2024-07-02 12:26:37,202 [salt.loaded.int.module.cmdmod:922 ][ERROR   ][30012] stdout: OpenSCAP Error: Unable to open file: '/root/scap/content/openscap/ssg-centos9-ds.xml' [/builddir/build/BUILD/openscap-1.3.10/src/source/oscap_source.c:298]
~~~

(need the CentOS Stream 9 content in the "massaged" content-bundle).

ash-linux/el9/RuleById/medium/content_rule_mount_option_tmp_noexec.sls
lorengordon: Ahh, getting back to "this setup is hard to understand," I had missed the detail about `OptsDefaultFile` vs `optionsFile`. I'll refine this some more with that in mind. Gimme a bit...

ferricoxide: No worries. It was a file that kind of grew (basically, it started with "just create the override file with just the STIG mount-opts" until, after a couple runs, I realized, "shit, I _need_ the mount-opts from the `/usr` file, too).

ash-linux/el9/VendorSTIG/remediate.sls
ferricoxide: Moved. Testing, nao.

ferricoxide: Ok, I _believe_ the move was non-fatal. Post execution with the new location:

```
[root@ip-bnosvxdsvvx ~]# chage -l root
Last password change                                    : Feb 07, 2024
Password expires                                        : never
Password inactive                                       : never
Account expires                                         : never
Minimum number of days between password change          : -1
Maximum number of days between password change          : -1
Number of days of warning before password expires       : 7
```

ash-linux/el9/RuleById/medium/content_rule_mount_option_tmp_noexec.sls
lorengordon: Ok, so this becomes much easier if we switch to just managing the entire `optionsFile`... Just unique and join the list of options!

```
Manage {{ optionsFile }} ({{ stig_id }}):
  file.managed:
    - name: '{{ optionsFile }}'
    - contents: |
        [Mount]
        Options={{ mountOptsDefault + mountOptsStig | unique | join(',') }}
    - dir_mode: '0755'
    - group: 'root'
    - makedirs: True
    - mode: '0644'
    - requires:
      - file: Create {{ optionsDir }} ({{ stig_id }})
    - user: 'root'
```

Does this also need the selinux options?

ferricoxide: Any file that gets laid down needs SELinux options.

ash-linux/el9/VendorSTIG/packages.sls
lorengordon: You don't really need to run it all. On the centos stream box, just run `salt-call --local grains.get os` and it will give you the value it sees for the grain. Then we'll know whether this test will ever match. If not, can just remove it.

ash-linux/el9/RuleById/medium/content_rule_mount_option_tmp_noexec.sls
lorengordon: Right, ok, seeing this suggestion still seems relevant, removing the `file.replace` stanzas, and just using `file.manage` for the whole content.

.github/workflows/test.yml
lorengordon, line 1: Update the `os_version` list, removing `7` and adding `9`. That should get the tests running.

ash-linux/el9/RuleById/medium/content_rule_mount_option_tmp_noexec.sls
ferricoxide: Had to make a slight mod to the [suggested content](https://github.com/plus3it/ash-linux-formula/pull/506#discussion_r1662810726), but otherwise eliminated the "extra" stanzas.

ash-linux/el9/VendorSTIG/packages.sls
ferricoxide: Ok, so:

```bash
# salt-call --local grains.get os
local:
    CentOS Stream
```
Waiting for the rest of the userData to run to see if it bombs-out (I know on previous runs, it _had_ &ndash; why I'd changed to testing _only_ on RHEL 9 &ndash; but don't recall the specific failure)

ferricoxide: Ah. Yes. That's right:

~~~
OpenSCAP Error: Unable to open file: '/root/scap/content/openscap/ssg-centos9-ds.xml' [/builddir/build/BUILD/openscap-1.3.10/src/source/oscap_source.c:298]
~~~

_Can't_ run on centos till watchmaker installs `ssg-centos9-ds.xml` (need to update the logic so it tries to run `…/ssg-cs9-ds.xml`, instead). In turn, can't tell if the conditional needs to be updated to something like:

~~~yaml
{%- if 'CentOS' in salt.grains.get('os')  %}
~~~

Or if that CentOS-specific RPM is no longer needed.

.github/workflows/test.yml
ferricoxide: Done

ash-linux/el9/VendorSTIG/packages.sls
ferricoxide: When I hard-code the path on the test-EC2, it gets past the `Unable to open file` error and doesn't fail due the lack of the `openscap-engine-sce` RPM, so, I'll remove that block.

ferricoxide: Opened [watchmaker-salt-content #44](https://github.com/plus3it/watchmaker-salt-content/issues/44) to address the problem coming from the default pillar content

ash-linux/el9/VendorSTIG/init.sls
```diff
@@ -0,0 +1,17 @@
...
+# do not yet (or will never) have suitable DISA- or OSCAP-provied
+# content - will be located in this code-branch
+#
+######################################################################
+
+include:
+  - ash-linux.el8.VendorSTIG.packages
+  - ash-linux.el8.VendorSTIG.remediate
+  - ash-linux.el8.VendorSTIG.scap_undos
+  - ash-linux.el8.VendorSTIG.aws_cli_v2
```
lorengordon, line 17: These el8 references should be `el9`.

```suggestion
  - ash-linux.el9.VendorSTIG.packages
  - ash-linux.el9.VendorSTIG.remediate
  - ash-linux.el9.VendorSTIG.scap_undos
  - ash-linux.el9.VendorSTIG.aws_cli_v2
```

ash-linux/el9/VendorSTIG/packages.sls
```diff
@@ -0,0 +1,14 @@
+# This Salt state downloads the tools necessary to scan,
+# remediate and report on the compliance-state of an EL8-based
```
lorengordon, line 2:
```suggestion
# remediate and report on the compliance-state of an EL9-based
```

ash-linux/el9/VendorSTIG/remediate.sls
```diff
@@ -0,0 +1,49 @@
+# This Salt state downloads the tools necessary to scan,
+# remediate and report on the compliance-state of an EL8-based
```
lorengordon, line 2:
```suggestion
# remediate and report on the compliance-state of an EL9-based
```

ash-linux/el9/VendorSTIG/report.sls
```diff
@@ -0,0 +1,36 @@
+# This Salt state downloads the tools necessary to scan,
+# remediate and report on the compliance-state of an EL8-based
```
lorengordon, line 2:
```suggestion
# remediate and report on the compliance-state of an EL9-based
```

ash-linux/el9/STIGbyID/cat2/RHEL-09-255065.sls
```diff
@@ -0,0 +1,53 @@
...
+# Finding ID: V-257989
+# Rule ID:    SV-257989r943014_rule
+# STIG ID:    RHEL-09-255065
+# SRG ID:     SRG-OS-000250-GPOS-00093
+#
+# Finding Level: medium
+#
+# Rule Summary:
+#       RHEL 9 must implement DOD-approved encryption ciphers to protect the
+#       confidentiality of SSH client connections.
```
eemperor, line 11: A little nit... 😃 
```suggestion
#       confidentiality of SSH server connections.
```

ash-linux/el9/RuleById/medium/content_rule_configure_tmux_lock_after_time.sls
```diff
@@ -0,0 +1,66 @@
...
+#     - FTA_SSL.1
+#   - OS-SRG
+#     - SRG-OS-000029-GPOS-00010
+#     - SRG-OS-000031-GPOS-00012
+#
+#################################################################
+{%- set stig_id = 'configure_tmux_lock_after_time' %}
+{%- set helperLoc = tpldir ~ '/files' %}
+{%- set skipIt = salt.pillar.get('ash-linux:lookup:skip-stigs', []) %}
+{%- set lockAfterSec = '900' %}
```
eemperor, line 28: Should we make this configurable from pillar as well?

Something like this?
```suggestion
{%- set lockAfterSec = salt.pillar.get('ash-linux:lookup:tmux-lock-after-time', '900') %}
```

ferricoxide: A system is never going to pass a scan if the value is greater than `900`, few users are going to want to make idle-locking _more_ aggressive  and, finally, some of the scanners only look for `900`. But, for completeness/flexibility sake, committing it "just in case".

eemperor (approved):

PR #591, opened 2025-12-04
Ensure the password complexity module in the system-auth file is configured for three retries or less.

### LINKED ISSUE #573: [Feature Request] Compliance Setting: "OS must ensure the password complexity module in the system-auth file is configured for three retries or less"

STIG scans calling out RHEL-09-611010/OL09-00-001001/ALMA-09-035990

Per the STIG guidance:

> Configure the OS to limit the "`pwquality`" retry option to "`3`". Add the following line to the "`/etc/pam.d/system-auth`" and "`/etc/pam.d/password-auth`" files (or modify the line to have the required value):
> ```
> password required pam_pwquality.so retry=3
> ```

**Additional context**

ferricoxide:
Closes #573

### DIFF
--- a/ash-linux/el9/STIGbyID/cat2/RHEL-09-611010.sls
+++ b/ash-linux/el9/STIGbyID/cat2/RHEL-09-611010.sls
@@ -75,19 +75,16 @@ notify_{{ stig_id }}-skipSet:
 Modify {{ pwqualityCfgFile }}:
   file.replace:
     - name: '{{ pwqualityCfgFile }}'
-    - pattern: '(^(|#)(|\s\s*)retry)((|\s\s*)=(|\s\s*))\d'
+    - append_if_not_found: True
+    - not_found_content: |
+
+        # Set per rule {{ stig_id }}
+        # Prompt user at most '{{ retryTimes }}' times before returning with error
+        retry = {{ retryTimes }}
+    - onlyif:
+      - 'grep -qP "retry(|\s\s*)=" {{ pwqualityCfgFile }}'
+    - pattern: '(^(|\s\s*)retry)((|\s\s*)=(|\s\s*))\d'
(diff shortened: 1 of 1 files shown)


ash-linux/el9/STIGbyID/cat2/RHEL-09-611010.sls
```diff
@@ -0,0 +1,90 @@
...
+Modify {{ pwqualityCfgFile }}:
+  file.replace:
+    - name: '{{ pwqualityCfgFile }}'
+    - append_if_not_found: True
+    - not_found_content: |
+
+        # Set per rule {{ stig_id }}
+        # Prompt user at most '{{ retryTimes }}' times before returning with error
+        retry = {{ retryTimes }}
+    - onlyif:
```
eemperor: Why do you need `onlyif` here?  I'm reading this as the state won't run if `retry =` is not found.  Don't you want the state to run when `retry =` is not found?

ferricoxide: Given my `salt.file.find` action's `grep` conent, it's _probably_ redundant, but I was mostly trying to (belt-and-suspenders my way to) only having the `append_if_not_found` or `repl` content added if a file had `retry` specified _anywhere_. However, I figured that, unlike the `/etc/security/pwquality.conf`,  there wouldn't be any `/etc/security/pwquality.conf.d/*.conf` files that had it commented out. So, basically, the `append_if_not_found` would only be applied to the `/etc/security/pwquality.conf` and the `repl` content would only be applied to `/etc/security/pwquality.conf.d/*.conf` files or the `/etc/security/pwquality.conf` file if someone/something had already placed _live_ `retry` content within them.

Probably overkill, but, having tested it, I know that it works as intended.

PR #636, opened 2026-06-02
Adds al2023 mappings to all el9 stig states

lorengordon:


ferricoxide (commented):
Web interface makes this hard to sort out. Broadly speaking a given platform's mapping should go to the platform-specific ID for the finding being remediated. If what the state-file is designed to remediate, then a platform-block like:

```yaml
{%- elif (
    osName != 'AlmaLinux' and
    osName != 'Amazon'
  )
%}
```
or:
```yaml
{%- elif osName == 'AlmaLinux' %}
```

Should be included that causes the remediation-logic to be skipped for the platform.  I'm not seeing that in this PR and, without cross-referencing to the stig-viewer, can't tell if that content is (effectively) "missing".

Basically, don't want to apply a RHEL-, ALMA- or OL-specific remediations to AMZN targets.

ferricoxide (approved):
Have created a ticket for adding appropriate execution-gating to relevant files. Approving this ticket "as is"

